PR #1, opened 2014-11-10
Refactoring and cleaning of the S3 outputs

ph:
This PRs includes a lot of changes.

- Merged @jsvd documentation improvement and added more documentation on new attributes Fix: https://github.com/elasticsearch/logstash/issues/1340
- Ruby code cleanup, following the STYLE.md
- It now use the aws_config mixing support IAM roles fix https://github.com/elasticsearch/logstash/issues/1778 fix https://github.com/elasticsearch/logstash/issues/1575
- the plugin also use standard codec, added a new codec [s3_plain](https://github.com/logstash-plugins/logstash-codec-s3plain) to support previous version
- refactoring of the code to improve clarity and allow easier testing
- Created initial test suite for it.
- Fixed a bug causing events to be dropped when using the size_file option and the file rotation.
- Fixed a bug when the tags were nil in the plain text format. fix https://github.com/elasticsearch/logstash/issues/1626
- Fixed a bug when specifying the file_size, the doc says KB but the condition was on bytes.
- Added the option to specify a temporary_directory for the working files. fix https://github.com/elasticsearch/logstash/issues/1860
- S3 output issue when working with multiple workers added a threaded uploader fix https://github.com/elasticsearch/logstash/issues/1915
- S3 output: when the codec is started it will try to create a simple file on your bucket to test the permissions fix https://github.com/elasticsearch/logstash/issues/1914

### DIFF
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -112,10 +112,10 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   # Specify how many workers to use to upload the files to S3
   config :upload_workers_count, :validate => :number, :default => 1
 
-
   # Exposed attributes for testing purpose.
   attr_accessor :tempfile
   attr_reader :page_counter
+  attr_reader :s3
 
   def aws_s3_config
     @logger.info("Registering s3 output", :bucket => @bucket, :endpoint_region => @region)
@@ -233,7 +233,6 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
       when LogStash::ShutdownEvent
         @logger.debug("S3: upload worker is shutting down gracefuly")
         @upload_queue.enq(LogStash::ShutdownEvent)
-        break
       else
         @logger.debug("S3: upload working is uploading a new file", :filename => File.basename(file))
         move_file_to_bucket(file)
@@ -263,8 +262,26 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
 
     begin
       write_on_bucket(test_filename)
+      delete_on_bucket(test_filename)
     ensure
       File.delete(test_filename)
+    end
+  end
+  
+  def delete_on_bucket(filename)
+    bucket = @s3.buckets[@bucket]
+
+    remote_filename = "#{@prefix}#{File.basename(filename)}"
+
+    @logger.debug("S3: delete file from bucket", :remote_filename => remote_filename, :bucket => @bucket)
+
+    begin
+      # prepare for write the file
+      object = bucket.objects[remote_filename]
+      object.delete
+    rescue AWS::Errors::Base => e
+      @logger.error("S3: AWS error", :error => e)
+      raise LogStash::ConfigurationError, "AWS Configuration Error"
     end
   end
 
@@ -306,16 +323,11 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
     @periodic_upload_thread = Thread.new do
       LogStash::Util::set_thread_name("<S3 periodic uploader")
 
-      first_interval = true
-      Stud.interval(@time_file * 60) do
-        if first_interval == false
-          @logger.debug("S3: time_file triggered, bucketing the file")
+      Stud.interval(@time_file * 60, :sleep_then_run => true) do
+        @logger.debug("S3: time_file triggered, bucketing the file")
 
-          move_file_to_bucket_async(@tempfile.path)
-          create_temporary_file
-        else
-          first_interval = false
-        end
+        move_file_to_bucket_async(@tempfile.path)
+        create_temporary_file
       end
     end
   end
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -148,9 +148,9 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
       # prepare for write the file
       object = bucket.objects[remote_filename]
       object.write(:file => file, :acl => @canned_acl)
-    rescue AWS::Errors::Base => e
-      @logger.error("S3: AWS error", :error => e)
-      raise LogStash::Error, "AWS Configuration Error"
+    rescue AWS::Errors::Base => error
+      @logger.error("S3: AWS error", :error => error)
+      raise LogStash::Error, "AWS Configuration Error, #{error}"
     end
 
     @logger.debug("S3: has written remote file in bucket with canned ACL", :remote_filename => remote_filename, :bucket  => @bucket, :canned_acl => @canned_acl)
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -207,7 +207,151 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
     end
   end
 
+
+  # Use the same method that Amazon use to check
+  # permission on the user bucket by creating a small file
   public
+  def test_s3_write
+    @logger.debug("S3: Creating a test file on S3")
+
+    test_filename = File.join(@temporary_directory,
+                              "logstash-programmatic-access-test-object-#{Time.now.to_i}")
+
+    File.open(test_filename, 'a') do |file|
+      file.write('test')
+    end
+
+    begin
+      write_on_bucket(test_filename)
+      delete_on_bucket(test_filename)
+    ensure
+      File.delete(test_filename)
+    end
+  end
+  
+  public
+  def restore_from_crashes
+    @logger.debug("S3: is attempting to verify previous crashes...")
+
+    Dir[File.join(@temporary_directory, "*.#{TEMPFILE_EXTENSION}")].each do |file|
+      name_file = File.basename(file)
+      @logger.warn("S3: have found temporary file the upload process crashed, uploading file to S3.", :filename => name_file)
+      move_file_to_bucket_async(file)
+    end
+  end
+
+  public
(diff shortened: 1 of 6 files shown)


luismfonseca:
Can anyone estimate when will this be merged?

ph:
@suyograo can you review this? It also have the feature describe in  #2

spec/outputs/s3_spec.rb
```diff
@@ -3,4 +3,238 @@
...
+  end
+
+  let(:minimal_settings)  {  { "access_key_id" => "1234",
+                               "secret_access_key" => "secret",
+                               "bucket" => "my-bucket" } }
+
+  describe "configuration" do
+    it "should support the deprecated endpoint_region as a configuration option" do
+      config = { "endpoint_region" => "sa-east-1" }
+      s3 = LogStash::Outputs::S3.new(config)
```
purbon: Just a matter of style, but this two lines could be generalized using using let statements.

spec/outputs/s3_spec.rb
```diff
@@ -3,4 +3,238 @@
...
+      expect(s3.aws_options_hash[:s3_endpoint]).to eq("s3-sa-east-1.amazonaws.com")
+    end
+
+    it "should use the depracated option before failling back to the region" do
+      config = { "region" => "us-east-1", "endpoint_region" => "sa-east-1" }
+      s3 = LogStash::Outputs::S3.new(config)
+      expect(s3.aws_options_hash[:s3_endpoint]).to eq("s3-sa-east-1.amazonaws.com")
```
purbon: To check that you can use this matcher http://www.relishapp.com/rspec/rspec-expectations/v/2-14/docs/built-in-matchers/include-matcher#hash-usage

spec/outputs/s3_spec.rb
```diff
@@ -3,4 +3,238 @@
...
+        s3.tempfile = tmp
+        expect(s3.rotate_events_log?).to be(true)
+      end
+    end
+
+    it "returns false if the tempfile is under the file_size limit" do
+      Stud::Temporary.file do |tmp|
+        tmp.stub(:size) { 100 }
+
+        s3 = LogStash::Outputs::S3.new(minimal_settings.merge({ "size_file" => 1024 }))
```
purbon: This might also be generalized into using a let.

spec/integration/s3_spec.rb
```diff
@@ -0,0 +1,100 @@
+require "logstash/devutils/rspec/spec_helper"
+require "logstash/outputs/s3"
+require 'socket'
+require "aws-sdk"
+require "fileutils"
+require "stud/temporary"
+
+describe LogStash::Outputs::S3, :integration => true, :s3 => true do
+  let!(:minimal_settings)  {  { "access_key_id" => ENV['AWS_ACCESS_KEY_ID'],
```
purbon: Can we add, for example in a simple readme, that for testing we need to create env variables like described here? It's important for example when we try to test all plugins together.

spec/integration/s3_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      s3output.register
+      s3output.s3
+  end
+
+  after(:all) do
+    delete_matching_keys_on_bucket('studtmp')
+    delete_matching_keys_on_bucket('my-prefix')
+  end
+
+  def delete_matching_keys_on_bucket(prefix)
```
purbon: you can "move" this functions into a support module, so they not mix with the test cases.

purbon:
As a general comment on the specs, I would enforce the idea of using more the power of subjects, something like what we can see http://www.relishapp.com/rspec/rspec-core/v/2-11/docs/subject/explicit-subject#subject-in-top-level-group, what do you think?

spec/integration/s3_spec.rb
ph: I have added a small readme for this plugin, but we need to complete it from the details of https://github.com/elasticsearch/dev/issues/471

purbon: sure, just aimed to have here a note, so if we test plugins we now we need to pass the details like this. :-)

ph:
@purbon I have added the tests to ensure we don't lose any events when doing the file rotation.
Concerning using more subjects, could you take one or 2 example in the suite to give me an idea how you would do it?

lib/logstash/outputs/s3.rb
```diff
@@ -73,285 +61,353 @@
...
-       end
-     end
+  # Use the same method that Amazon use to check
+  # permission on the user bucket by creating a small file
+  public
+  def test_s3_write
+    @logger.debug("S3: Creating a test file on S3")
 
-     File.delete (file)
+    test_filename = File.join(@temporary_directory, "logstash-programmatic-access-test-object")
```
suyograo: Do we need a timestamp for this temp file name?

ph: After thinking, this would make sure the check works if something goes wrong.
I'll add it.

ph: Not sure about the `verify_*`

spec/outputs/s3_spec.rb
```diff
@@ -1,6 +1,318 @@
...
+        expect(events_written_count).to eq(event_count)
+      end
+    end
+
+    it "doesn't skip events if using the time_file option", :tag => :slow do
+      Stud::Temporary.directory do |temporary_directory|
+        time_file = rand(5..10)
+        number_of_rotation = rand(4..10)
+
+        config = {
```
purbon: Just a cosmetic comment: can we make this test, or might be the other work the same way? or either both using the pipeline or just the output? Then we might be able to reduce the tests size by reusing some of the constructs using lets  in different contexts, like one context for using the size_file and another with the temp_file.

jsvd:
some of the methods marked as public can be converted to private like create_temporary_file configure_upload_workers next_page test_s3_write

spec/outputs/s3_spec.rb
ph: I tried using the pipeline for both, but for `Timed constraints` it wasn't possible or I don't know how and using the pipeline make sense for the first one. I will leave them like this.

ph:
@jsvd good catch I'll review them.

lib/logstash/outputs/s3.rb
```diff
@@ -73,285 +61,354 @@
...
+        @logger.debug("S3: tempfile is too large, let's bucket it and create new file", :tempfile => File.basename(@tempfile))
+
+        move_file_to_bucket_async(@tempfile.path)
+        next_page
+        create_temporary_file
+      else
+        @logger.debug("S3: tempfile file size report.", :tempfile_size => @tempfile.size, :size_file => @size_file)
+      end
+
+      write_to_tempfile(encoded_event)
```
jsvd: maybe move this to the outside of the if, since it will be called either way

spec/outputs/s3_spec.rb
purbon: Ok, for now is ok. they both test the problems.

ph:
adressed @jsvd and @purbon comments.

spec/outputs/s3_spec.rb
```diff
@@ -1,6 +1,322 @@
...
+
+      s3 = LogStash::Outputs::S3.new(minimal_settings)
+      allow(s3).to receive(:test_s3_write)
+      s3.register
+
+      s3.receive(event)
+    end
+  end
+
+  describe "when rotating the temporary file" do
```
purbon: As far as this test are using a real S3 for testing, they should go into the integration part.

lib/logstash/outputs/s3.rb
```diff
@@ -73,285 +61,354 @@
...
- # 0 stay all time on listerner, beware if you specific 0 and size_file 0, because you will not put the file on bucket,
- # for now the only thing this plugin can do is to put the file when logstash restart.
- config :time_file, :validate => :number, :default => 0
+    begin
+      # prepare for write the file
+      object = bucket.objects[remote_filename]
+      object.write(:file => file, :acl => @canned_acl)
+    rescue AWS::Errors::Base => e
+      @logger.error("S3: AWS error", :error => e)
+      raise LogStash::ConfigurationError, "AWS Configuration Error"
```
jsvd: Not sure if ConfigurationError is the right kind of exception to raise when a write operation fails...maybe LogStash::Error?

purbon: Definety not a configuration error.

ph: I will change it for LogStash::Error.

spec/outputs/s3_spec.rb
ph: Well they test if the file rotation is done locally and we don't drop events when creating new files. Not sure why this local test would require a real S3 connection?

purbon: It requires a local connection for me, you should probably mock the connection parts.

ph: I'll do more testing, the `AWS.stub!` at https://github.com/logstash-plugins/logstash-output-s3/pull/1/files#diff-14cf776b93312a411664cdaf9463b8b3R14 should mock all the connections made from the aws-sdk.

purbon: not in this case, at less for me.

ph: Adding a note:
The call was mocked, but the S3 settings were not explicitly set in the config for the test.
The `aws_mixin` have multiples fallback and it was picking the config in my environment.

PR #12, opened 2015-03-30
Allow serverside encryption as an option.

rfkrocktk:
A simple true/false toggle has been added as :serverside_encryption for whether to use Amazon's AES-256 serverside encryption. I have tested with it on and off and it works as expected.


suyograo:
Fixed in #62

Issue #3, opened 2014-12-22
Support file compression before uploading to s3


adamvduke:
It would be valuable to have the ability to configure the plugin to compress files before uploading them to s3. At the moment we manually download, compress, upload, delete uncompressed files. Perhaps something like:

```
s3  {
  ...
  compress => 'gzip'
}
```

This is a feature I'm considering implementing, but I wanted to get a sense of if the maintainers would find this valuable, and/or if there are any other concerns I should be aware of before taking it on.

I see that https://github.com/logstash-plugins/logstash-output-s3/pull/1 is a fairly extensive refactor. Is there a timeline for when that is expected to be merged? I'd like to avoid having to rebase against a large change.

oroce:
Any update on this? Is it decided whether gzip should be a codec or not?
I have a working solution for 1.4: https://github.com/oroce/logstash-output-s3/blob/gzip-configurable-dir/lib/logstash/outputs/s3.rb#L378-L385
But I would love to upgrade to 1.5 and even add gzip support to this module.

navatm:
@oroce I was not able to use your code.
In debug mode I got
S3: file: ls.s3.erans-mini-2.2015-04-16T11.51.tag_s3.part0.txt was put on bucket test-logstash {:level=>:debug, :file=>"logstash/outputs/s3.rb", :line=>"250"}
Without any file in my S3 bucket.
The local temporary file remained empty as well.

oroce:
@navatm hm i was using that module for like 2 months with the following configuration without any problems:
~~~
output {
	stdout { codec => rubydebug }
	s3 {
		bucket => '<bucket>'
		access_key_id => '<access>'
		endpoint_region => 'eu-west-1'
		format => 'json'
		secret_access_key => '<secret>'
		#time_file => 5
		size_file => 504800
		gzip => true
	}
}
~~~

Sadly I'm not using it anymore, because I implemented it just for one of my sideprojects.

shribigb:
@ph When this is going to be merged? I would like to have this compression support.

navatm:
@shribigb we've been working with https://github.com/adience-code/logstash-output-s3 on production for a month. No error yet any problems, ~100GB a day, compress and flush every 4 hours.

shaharmor:
+1

navatm:
Since your name is familiar, I just added a pull request.

vpack:
+1

parangdae:
+1

[...]

PR #18, opened 2015-06-01
Add gzip support for #3

a86c6f7964:


a86c6f7964:
It doesn't look like this got linked to #3 correctly

shribigb:
I gave this pull request a try and I noticed that its missing few seconds of data. Also most of the time last message is  truncated at the end of the file.

a86c6f7964:
Is that just a general problem? I am not really sure why it would be the case given that its just wrapping an existing pipe

a86c6f7964:
i am seeing that behavior as well. i wonder if there is a missing flush that was still working because of the syswrite call

shribigb:
That last fix seems to fix the problem. Now I dont see missing log lines as well as truncated files. Thanks a lot.

a86c6f7964:
Great. Works well for me too

suyograo:
@a86c6f7964 this looks great! can you please add tests to validate this behavior?

a86c6f7964:
@suyograo I do not know much about testing ruby stuff. maybe if someone else can add tests? it is mostly tested by any of the existing logical tests since all it does is wrap the gzip stream around the file stream that already exists.

suyograo:
@a86c6f7964 sorry about the delays here.. can you rebase please?

a86c6f7964:
okay

a86c6f7964:
I rebased it, fixed the conflicts, and it looks like its passing the tests.

a86c6f7964:
does my rebase work?

Issue #15, opened 2015-04-24
IOError: Connection reset by peer


calve:
Hello.
On my configuration (Ubuntu 14.04, Logstash 1.4), ``logstash-output-s3`` randomly crash with no obvious reason.
This bug is really annoying as it crash logstash too, and thus, logs sent by hosts are lost by logstash.

Here is the stacktrace I have after the plugin crash. Please, let me know if you need more informations

```
IOError: Connection reset by peer
                       syswrite at org/jruby/ext/openssl/SSLSocket.java:667
                       do_write at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/shared/jopenssl19/openssl/buffering.rb:318
                          write at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/shared/jopenssl19/openssl/buffering.rb:336
                         write0 at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/1.9/net/protocol.rb:199
                          write at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/1.9/net/protocol.rb:173
                        writing at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/1.9/net/protocol.rb:190
                          write at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/1.9/net/protocol.rb:172
  send_request_with_body_stream at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/1.9/net/http.rb:1963
                           exec at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/1.9/net/http.rb:1929
              transport_request at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/1.9/net/http.rb:1325
                          catch at org/jruby/RubyKernel.java:1284
              transport_request at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/1.9/net/http.rb:1324
[...]

PR #20, opened 2015-06-25
Fix a synchronization issue when doing file rotation

ph:
Added a `synchronize` block when trying to fetch the current size of the
temp file, this method could have been call when doing a file rotation
and throw `OError: Connection reset by peer` Fixes: #15


ph:
@purbon Can you look at this?

purbon:
@ph looking at this.

purbon:
@ph code looks good, can we add a test to cover this situation?

ph:
@purbon Its a race condition how would you test this?

purbon:
@ph can you describe the condition, then we might be able to workout  a test. I mean, I guess you did that fix, because you were somehow able to reproduce the condition, isn't? which was your process?

ph:
The condition was that two different threads was competing to access the same File handle.
One of the thread was doing a periodic rotation of the file and the writing thread was looking at the size of the file to know if we need to rotate it. 

Since the file was being 'moved' the writer would throw a `OError: Connection reset by peer` all the other access to the file were alreading using a shared lock except the *file size check*, This PR add the lock to this method.

This is a race condition and it isa  bit hard to reproduce in a local environment, Some people tried the patch and it fixed their problem.

ph:
If you want any idea how we could add a test for that behavior I would really like to know how to do it.

ph:
edited last comment since it didnt make any sense :)

purbon:
you could see an initial proposal of test at https://github.com/ph/logstash-output-s3/pull/1 , not perfect, but already more than not having anything ... looking forward to iterate them :P

ph:
@purbon I have added your specs and it LGTM and I will release it. TIL.

purbon:
@ph Cool, I know they are not perfect, but at less is something. Keep in mind to rebase the commits, especially the merge commits, looks not clever to have in the history, don't you think?

ph:
dont worry I'll squash it.

PR #33, opened 2015-09-21
Support for dynamic prefix field based s3 directory separation

nigoel:
One may not provide the dynamic prefixes like "logs/%{app}/%{type}/" . Fields can be extracted from the event messages.

This will create the temp file locally per prefix and apply a time_file watch on each prefix. Change also maintains a separate file lock for each prefix.

After no_event_wait time, it will reset the watch and do a local clean-up for the files under a prefix. This will happen if there has been no event for that prefix, in last no_event_wait * time_file time from the last event.

### DIFF
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -8,7 +8,7 @@ require "socket" # for Socket.gethostname
 require "thread"
 require "tmpdir"
 require "fileutils"
-require 'pathname'
+require "pathname"
 
 
 # INFORMATION:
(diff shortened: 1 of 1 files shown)


nigoel:
Can we run the CLA check again

jordansissel:
@nigoel just re-ran it. Still failing.

Is your git commit author (see the `From:` line on [the patch](https://patch-diff.githubusercontent.com/raw/logstash-plugins/logstash-output-s3/pull/33.patch), or your `git log`) the same as the one you signed the CLA under?

nigoel:
@jordansissel Please re-run.  I have signed it with the mail id you mentioned. However the public id on my github is different.

PR #35, opened 2015-09-23
rename teardown to close and fix dependencies

### LINKED ISSUE #34: Test hanging because of the change to use Stud::Task
When I did the refactor for the new shutdown semantic I replaced the `Thread.new` with `Stud::Task` and this cause the test to hang.

jsvd:
fixes #34


ph:
LGTM

PR #38, opened 2015-09-30
Fixed a small typo

Dawny33:


lib/logstash/outputs/s3.rb
```diff
@@ -13,7 +13,7 @@
 # INFORMATION:
 #
 # This plugin sends logstash events to Amazon Simple Storage Service (Amazon S3).
-# To use it you need to have the proper write premissions and a valid s3 bucket.
```
untergeek: Thank you for taking the time to submit this change, but it is unnecessary.  "To use it" is proper english.

PR #41, opened 2015-10-13
Reintroduce `tags` in the plugin

### LINKED ISSUE #39: specs still use "tags", which are obsolete in 2.0
failure in defaults plugins test: http://build-eu-00.elastic.co/view/LS%20Master/job/Logstash_Master_Default_Plugins/612/console

```
  1) LogStash::Outputs::S3#generate_temporary_filename should not add the tags to the filename
     Failure/Error: s3 = LogStash::Outputs::S3.new(config)
     LogStash::ConfigurationError:
       The setting `tags` in plugin `s3` is obsolete and is no longer available. You can achieve similar behavior with the new conditionals, like: `if "sometag" in [tags] { s3 { ... } }` If you have any questions about this, you are invited to visit https://discuss.elastic.co/c/logstash and ask.
     # ./lib/logstash/config/mixin.rb:82:in `config_init'
     # ./lib/logstash/config/mixin.rb:66:in `config_init'
     # ./lib/logstash/outputs/base.rb:44:in `initialize'
     # ./vendor/bundle/jruby/1.9/gems/logstash-output-s3-2.0.1/spec/outputs/s3_spec.rb:84:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.7/lib/rspec/wait.rb:46:in `(root)'
     # ./rakelib/test.rake:47:in `(root)'
```

Code is: https://github.com/logstash-plugins/logstash-output-s3/blob/master/spec/outputs/s3_spec.rb#L76-L86

ph:
Logstash 2.0 has marked `tags` as an obsolete option for filtering the
events. But this plugins also uses tags for generating the filename
before sending it to s3.

Fixes #39

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 2.0.2
+ - Fixes an issue when tags were defined #39
 ## 2.0.0
  - Plugins were updated to follow the new shutdown semantic, this mainly allows Logstash to instruct input plugins to terminate gracefully, 
    instead of using Thread.raise on the plugins' threads. Ref: https://github.com/elastic/logstash/pull/3895
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -110,6 +110,16 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   # Specify how many workers to use to upload the files to S3
   config :upload_workers_count, :validate => :number, :default => 1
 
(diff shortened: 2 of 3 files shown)


suyograo:
@ph why do we need tags here?

suyograo:
ah nm, I see the usage.

suyograo:
LGTM

ph:
@suyograo `Tags` were not only used for filtering the events but also for generating the filename.

PR #42, opened 2015-10-21
add content encoding option

naivefun:
When content encoding is missing some non ascii characters is not human
readable from S3 browser


naivefun:
CAL is signed now, thanks.

kares:
should no longer be necessary, also this potentially clashes with the `encoding => gzip` option

PR #44, opened 2015-10-28
Support for dynamic prefix field based s3 directory separation

elyscape:
Based on #33, but rebased and with author's email set to the personal one listed on his GitHub profile, which is likely the address he used to sign the CLA.

From #33's description:

>One may not provide the dynamic prefixes like "logs/%{app}/%{type}/" . Fields can be extracted from the event messages.
>
>This will create the temp file locally per prefix and apply a time_file watch on each prefix. Change also maintains a separate file lock for each prefix.
>
>After no_event_wait time, it will reset the watch and do a local clean-up for the files under a prefix. This will happen if there has been no event for that prefix, in last no_event_wait * time_file time from the last event.

Closes #33.

### DIFF
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -482,8 +482,13 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   private
   def delete_on_bucket(filename)
     bucket = @s3.buckets[@bucket]
+    
+    first = Pathname.new @temporary_directory
+    second = Pathname.new filename
 
-    remote_filename = "#{@prefix}#{File.basename(filename)}"
+    remote_filename_path = second.relative_path_from first
+    
+    remote_filename = remote_filename_path.to_s
 
     @logger.debug("S3: delete file from bucket", :remote_filename => remote_filename, :bucket => @bucket)
 
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -8,6 +8,7 @@ require "socket" # for Socket.gethostname
 require "thread"
 require "tmpdir"
 require "fileutils"
+require 'pathname'
 
 
 # INFORMATION:
@@ -60,6 +61,7 @@ require "fileutils"
 #      time_file => 5                           (optional)
 #      format => "plain"                        (optional)
 #      canned_acl => "private"                  (optional. Options are "private", "public_read", "public_read_write", "authenticated_read". Defaults to "private" )
+#      no_event_wait => 5                       (optional. Defines the number of time_file s3 upload events that may go with no eventns for the prefix, before cleaning up the watch on that)       
 #    }
 #
 class LogStash::Outputs::S3 < LogStash::Outputs::Base
@@ -110,6 +112,9 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   # Specify how many workers to use to upload the files to S3
   config :upload_workers_count, :validate => :number, :default => 1
 
+  # Specify after how many interval of time_file, a prefix directory should be cleaned up locally if no events happing for it
+  config :no_event_wait, :validate => :number, :default => 5
+
   # Define tags to be appended to the file on the S3 bucket.
   #
   # Example:
@@ -149,8 +154,13 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   def write_on_bucket(file)
     # find and use the bucket
     bucket = @s3.buckets[@bucket]
+    
+    first = Pathname.new @temporary_directory
+    second = Pathname.new file
 
-    remote_filename = "#{@prefix}#{File.basename(file)}"
+    remote_filename_path = second.relative_path_from first
+    
+    remote_filename = remote_filename_path.to_s
 
     @logger.debug("S3: ready to write file in bucket", :remote_filename => remote_filename, :bucket => @bucket)
 
@@ -170,17 +180,21 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
 
   # This method is used for create new empty temporary files for use. Flag is needed for indicate new subsection time_file.
   public
-  def create_temporary_file
-    filename = File.join(@temporary_directory, get_temporary_filename(@page_counter))
-
-    @logger.debug("S3: Creating a new temporary file", :filename => filename)
-
-    @file_rotation_lock.synchronize do
-      unless @tempfile.nil?
-        @tempfile.close
+  def create_temporary_file(prefix)
+    filename = File.join(@temporary_directory, prefix, get_temporary_filename(@page_counter[prefix]))
+    @file_rotation_lock[prefix].synchronize do
+      unless @tempfile[prefix].nil?
+        @tempfile[prefix].close
       end
-
-      @tempfile = File.open(filename, "a")
+   
+      if @prefixes.include? prefix
+         dirname = File.dirname(filename)
+         unless File.directory?(dirname)
+           FileUtils.mkdir_p(dirname)
+         end
+         @logger.debug("S3: Creating a new temporary file", :filename => filename)
+         @tempfile[prefix] = File.open(filename, "a")
+      end
     end
   end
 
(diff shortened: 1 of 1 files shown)


nigoel:
Thanks @elyscape . I was struggling with the CLA. Thanks for updating it.

sijis:
I'm really looking forward to this.

suyograo:
@elyscape can you please add tests to validate these changes?

elyscape:
I'll take a look. @nigoel, if you can help on this some, that would be very appreciated.

lib/logstash/outputs/s3.rb
```diff
@@ -8,6 +8,7 @@
 require "thread"
 require "tmpdir"
 require "fileutils"
+require 'pathname'
```
ph: Lets use the same style as the other requires

require "pathname"

lib/logstash/outputs/s3.rb
```diff
@@ -60,6 +61,7 @@
 #      time_file => 5                           (optional)
 #      format => "plain"                        (optional)
 #      canned_acl => "private"                  (optional. Options are "private", "public_read", "public_read_write", "authenticated_read". Defaults to "private" )
+#      no_event_wait => 5                       (optional. Defines the number of time_file s3 upload events that may go with no eventns for the prefix, before cleaning up the watch on that)
```
ph: small typo in the doc, `eventns` should be `events`

lib/logstash/outputs/s3.rb
```diff
@@ -149,8 +154,13 @@ def aws_service_endpoint(region)
   def write_on_bucket(file)
     # find and use the bucket
     bucket = @s3.buckets[@bucket]
+    
+    first = Pathname.new @temporary_directory
```
ph: lets use parentheses for methods calls.  

```ruby
Pathname.new(@temporary_directory)
```

lib/logstash/outputs/s3.rb
```diff
@@ -170,17 +180,21 @@ def write_on_bucket(file)
...
-      unless @tempfile.nil?
-        @tempfile.close
+  def create_temporary_file(prefix)
+    filename = File.join(@temporary_directory, prefix, get_temporary_filename(@page_counter[prefix]))
+    @file_rotation_lock[prefix].synchronize do
+      unless @tempfile[prefix].nil?
+        @tempfile[prefix].close
+      end
+   
+      if @prefixes.include? prefix
```
ph: parentheses `()` for `include?`

lib/logstash/outputs/s3.rb
```diff
@@ -170,17 +180,21 @@ def write_on_bucket(file)
...
+         dirname = File.dirname(filename)
+         unless File.directory?(dirname)
+           FileUtils.mkdir_p(dirname)
+         end
```
ph: lets use a one liner here for unless

```ruby
FileUtils.mkdir_p(dirname) unless File.directory?(dirname)

**Comment by ph** [comment]
@elyscape @nigoel this will take a bit more time, I need to check if there any thread safety problems.

File: lib/logstash/outputs/s3.rb [hunk 6]
```diff
@@ -207,15 +225,14 @@ def register
     end
 
     test_s3_write
-
     restore_from_crashes if @restore == true
-    reset_page_counter
-    create_temporary_file
+    #reset_page_counter
+    #create_temporary_file
```
ph: We can remove this.

lib/logstash/outputs/s3.rb
```diff
@@ -252,13 +269,36 @@ def restore_from_crashes
...
+
   public
   def move_file_to_bucket(file)
+    
+    @logger.debug("S3: moving to bucket ", :file => file)
+
+    basepath = Pathname.new @temporary_directory
+    dirname = Pathname.new File.dirname(file)
+    prefixpath = dirname.relative_path_from basepath
+    prefix = prefixpath.to_s
```
ph: I dont see any use of prefixpath other that to set `prefix`
we can replace this with 

```ruby
prefix = dirname.relative_path_from(basepath).to_s
```

lib/logstash/outputs/s3.rb
```diff
@@ -322,9 +368,11 @@ def write_to_tempfile(event)
...
     @periodic_rotation_thread.stop! if @periodic_rotation_thread
-
-    @file_rotation_lock.synchronize do
-      @tempfile.close unless @tempfile.nil? && @tempfile.closed?
+    
+    for prefix in @prefixes
+       @file_rotation_lock[prefix].synchronize do
+         @tempfile[prefix].close unless @tempfile[prefix].nil? && @tempfile[prefix].closed?
+       end
     end
```
ph: We should use iterators in this case

```ruby
@prefixes.each do |prefix|
     @file_rotation_lock[prefix].synchronize do
        @tempfile[prefix].close unless @tempfile[prefix].nil? && @tempfile[prefix].closed?
      end
end

File: lib/logstash/outputs/s3.rb [hunk 9]
```diff
@@ -335,20 +383,29 @@ def shutdown_upload_workers
   end
 
   private
-  def handle_event(encoded_event)
+  def handle_event(encoded_event, event)
+    actualprefix = event.sprintf(@prefix)
+    if not @prefixes.to_a().include? actualprefix
```
ph: no need to convert the set into an array since **Set** class already support `include?`
```ruby
irb(main):002:0> require 'set'
=> true
irb(main):003:0> m = Set[1,2,3]
=> #<Set: {1, 2, 3}>
irb(main):004:0> m.include?(1)
=> true
irb(main):005:0>

Comment (ph) [inline lib/logstash/outputs/s3.rb]: We prefer not using `not` in conditionals but instead `!`

``ruby
if !@prefixes ...
```

lib/logstash/outputs/s3.rb
```diff
@@ -335,20 +383,29 @@ def shutdown_upload_workers
...
+       @file_rotation_lock[actualprefix] = Mutex.new
+       @prefixes.add(actualprefix)
+       reset_page_counter(actualprefix)
+       create_temporary_file(actualprefix)       
+       @empty_uploads[actualprefix] = 0       
+    end
```
ph: Might be worth it to move all this logic into his own method? What do you think?

`initialize_prefix(prefix)` or `configure_prefix(prefix)`

lib/logstash/outputs/s3.rb
```diff
@@ -357,16 +414,31 @@ def configure_periodic_rotation
...
+          move_file_to_bucket_async(@tempfile[key].path)
+          next_page(key)
+          create_temporary_file(key)
+        end
       end
     end
   end
 
   private
+  def cleanprefix(prefix)
```
ph: Can you rename `cleanprefix` for `clean_prefix`, we usually try to split words with underscore.

lib/logstash/outputs/s3.rb
```diff
@@ -252,13 +269,36 @@ def restore_from_crashes
     end
   end
 
+  public 
+  def shouldcleanup(prefix)
+     return @empty_uploads[prefix] > @no_event_wait
+  end
```
ph: Since this is a predicate method It should contains an integration mark.
I also suggest we rename it to `need_cleanup?(prefix)`  since it express more necessity.

lib/logstash/outputs/s3.rb
```diff
@@ -267,6 +307,10 @@ def move_file_to_bucket(file)
     rescue Errno::EACCES
       @logger.error("S3: Logstash doesnt have the permission to delete the file in the temporary directory.", :filename => File.basename(file), :temporary_directory => @temporary_directory)
     end
+
+    if shouldcleanup(prefix)
+        cleanprefix(prefix)
+    end
```
ph: A bit too much indentation, it could also be defined as a one liner
```ruby
clean_prefix(prefix) if need_cleanup?(prefix)

sijis:
Is there something I can do to get this moved on? Is it simply fixing the notes provided by @ph?

nigoel:
Thanks for the review. I will incorporate them.

nigoel:
@elyscape @ph I am not able to create a direct pull request to this because of CLA issues. 
I have created a pull request with changes incorporated in elyscape repo.@elyscape please help me here.

elyscape:
Sorry for the delay. Grabbed the changes from @nigoel and fixed them into this PR.

NoumanSaleem:
Would love to see this

PR #48, opened 2015-11-06
fix China s3 endpoint

josh-janrain:
Had to make this change in order to push to AWS S3 in China.


lib/logstash/outputs/s3.rb
```diff
@@ -140,7 +140,9 @@ def aws_service_endpoint(region)
     end
 
     return {
-      :s3_endpoint => region_to_use == 'us-east-1' ? 's3.amazonaws.com' : "s3-#{region_to_use}.amazonaws.com"
+        :s3_endpoint => region_to_use == 'us-east-1' ? 's3.amazonaws.com' :
+                        !(region_to_use =~ /^cn-.*/).nil? ? "s3.#{region_to_use}.amazonaws.com.cn" :
```
Janrain-Colin: The `.*` is redundant in this case, and can be removed.

Also, it looks like you used different indentation.

PR #53, opened 2015-11-16
Allow setting a header row to s3 files

coffeepac:
To support downstream consumption of files produced by logstash's s3 output, allow a hardcoded header row to be supplied in the configuration of the outputter.


coffeepac:
Closing to re-open under my username.

PR #54, opened 2015-11-16
Allow setting a header row to s3 files

coffeepac:
To support downstream consumption of files produced by logstash's s3 output, allow a hardcoded header row to be supplied in the configuration of the outputter.

### DIFF
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -133,6 +133,15 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   # Specify the content encoding. Supports ("gzip"). Defaults to "none"
   config :encoding, :validate => ["none", "gzip"], :default => "none"
 
+  # If set, this will be the first row of every file pushed to S3
+  #
+  # Ensure that you are setting this to the actual number of columns in your data.  It is not dynamic
+  # and will not help you out if the underlying schema changes.  
+  #
+  # Recommend to use wtih this a filter that produces an explicit set of fields no matter the input
+  # event
+  config :header_row, :validate => :string, :default => ""
+
   # Exposed attributes for testing purpose.
   attr_accessor :tempfile
   attr_reader :page_counter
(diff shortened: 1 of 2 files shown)


coffeepac:
@ph I hate to ping you, but I'm not sure what to do to get the CLA check to clear.  I've signed it twice and closed/reopened the PR after signing it.  Not sure what else I can do!  Any ideas?  Or should I be pinging someone else?

jordansissel:
I've confirmed the CLA has been signed. I don't know why the CLA check is failing.

lib/logstash/outputs/s3.rb
```diff
@@ -180,6 +189,11 @@ def create_temporary_file
       end
 
       @tempfile = File.open(filename, "a")
+      
+      #  header row processing
+      if @header_row != "" 
+        @tempfile.syswrite(@header_row + "\n")
+      end
```
ph: I assume we only want this to be added once in the file correct?

We are opening the file in `append` mode, if something weird happen to logstash and it has to restart we could have multiple headers added to the same file. This problem is rare since the restart has to happen in a one minute window. (see how we generate the filename https://github.com/logstash-plugins/logstash-output-s3/pull/54/files#diff-40972c051aeee75a31b999f62e7bcd65L279)

Might be worth it to check if there is already an header written to the file before adding a new one?

Also would you mind adding a test for this new feature?

ph:
@jordansissel @coffeepac The CLA check is failing because this patch was made with a `payscale.com` email and the the signature on record has a gmail address.

lib/logstash/outputs/s3.rb
coffeepac: Correct, once per file and only as the first line.  I'll add a check for filesize == 0 or something equivalent.

Will add a test or two, thanks for the reminder.

coffeepac:
@ph changes made and PR updated, can you take a look and see if its up to snuff?

thanks!

suyograo:
@coffeepac apologies for the delay here. Can you please rebase? i can merge then

coffeepac:
wow that was a really bad rebase.  going to take a bit to clean it up.  should be done today or tomorrow.

coffeepac:
@suyograo sorry for the delay on this, I've pulled my commit up to merge cleanly with current master.  Lemme know if I need to do anything else for this!  I've got a couple more improvements to share back as well.

jeyraof:
Is this issue working in progress?

coffeepac:
@jeyraof it looks like a pretty straightforward merge conflict.  I can do it if @suyograo or someone else on the logstash side is willing to accept it.

coffeepac:
attempted to rebase this again and the code has changed dramatically.  I don't need this anymore so going to abandon ship.

Issue #63, opened 2015-12-22
All default plugins test failing


ph:
http://build-eu-00.elasticsearch.org/view/LS%20Master/job/Logstash_Master_Default_Plugins/899/console

```
5) LogStash::Outputs::S3#generate_temporary_filename should add tags to the filename if present
     Failure/Error: expect(s3.get_temporary_filename).to eq("ls.s3.logstash.local.2015-01-01T00.00.tag_elasticsearch.logstash.kibana.part0.txt")
       
       expected: "ls.s3.logstash.local.2015-01-01T00.00.tag_elasticsearch.logstash.kibana.part0.txt"
            got: "ls.s3.logstash.local.2015-12-22T11.45.tag_elasticsearch.logstash.kibana.part0.txt"
       
       (compared using ==)
     # ./vendor/bundle/jruby/1.9/gems/logstash-output-s3-2.0.3/spec/outputs/s3_spec.rb:79:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.8/lib/rspec/wait.rb:46:in `(root)'
     # ./rakelib/test.rake:72:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/task.rb:240:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/task.rb:235:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/task.rb:179:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/task.rb:172:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/task.rb:165:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/application.rb:150:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/application.rb:115:in `run_with_threads'
[...]

PR #64, opened 2015-12-22
Remove the Time.now stubbing and make the test run faster.

ph:
Fixes: https://github.com/logstash-plugins/logstash-output-s3/issues/63


andrewvc:
LGTM

PR #69, opened 2016-02-02
Add server_side_encryption flag

### LINKED ISSUE #62: "server side" encryption
It would be nice to have an option to turn on [server side encryption](http://docs.aws.amazon.com/AmazonS3/latest/dev/serv-side-encryption.html) for S3 data

mattupstate:
I understand that there is a PR for this feature already, but I'm submitting this in an attempt to get this feature added a bit quicker given that I have accepted the CLA. As a result this would close #62 and #12.

### DIFF
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -99,6 +99,9 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   config :canned_acl, :validate => ["private", "public_read", "public_read_write", "authenticated_read"],
          :default => "private"
 
+  # Specifies wether or not to use S3's AES256 server side encryption. Defaults to false.
+  config :server_side_encryption, :validate => :boolean, :default => false
+
   # Set the directory where logstash will store the tmp files before sending it to S3
   # default to the current OS temporary directory in linux /tmp/logstash
   config :temporary_directory, :validate => :string, :default => File.join(Dir.tmpdir, "logstash")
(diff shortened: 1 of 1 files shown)


jtopper:
I'm keen to see this added too.

fatdrag0n:
Right now I have all logs systematcially very secure from our nodes to ELK server and we have it also stored on encrypted drive on aws instance. All of our data at this time, need to be transmitted securely in transit end-to-end and also at rest. S3 output plugin does communicate with use_ssl to s3 , which is great. but we need the s3 server side encryption request header to be requested from s3 output plugin.
We have three options from http://docs.aws.amazon.com/AmazonS3/latest/dev/serv-side-encryption.html and we are going with option1 , which I enabled the policy http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html for S3:PutObject to included header s3:x-amz-server-side-encryption":"AES256" and s3 output plugin is failing and is not sending our logs to s3.

{:timestamp=>"2016-04-19T22:36:12.171000+0000", :message=>"S3: AWS error", :error=>#<AWS::S3::Errors::AccessDenied: Access Denied>, :level=>:error}

If I delete the policy , it is transmitting the data but S3 saving without encryption!
Server side will encrypt our data before writing to disk and it will decrypt our data when we download from s3. We really need this feature. Thank you very much for your consideration.

fatdrag0n:
Does this mean, there is a fix? How do i download the fix?

MrPrimate:
@fatdrag0n Run the logstash plugin script for your version of logstash with `update log-stash-output-s3` command https://www.elastic.co/guide/en/logstash/current/working-with-plugins.html#updating-plugins

MrPrimate:
Would it be possible to get this merged into the 2.* stream? I'm unable to upgrade to 3.0.1 as I believe this requires the `logstash-core-plugin-api (~> 2.0)` which is a logstash 5 feature?

fatdrag0n:
Updated packages broke our configuration with encryption to s3 flags , it was working fine at 2.2.1 with manual update before and that's why I left it there..now it  LS version updated automatically 2.2.4 and encryption flag is broken!

jordansissel:
@MrPrimate This is available for Logstash 2.4.x and Logstash 5. I tested Logstash 2.4.1 myself and verified.

```
% bin/logstash --version
2.4.1
% bin/logstash-plugin install --version 3.2.0 logstash-output-s3
...
```

Issue #79, opened 2016-05-11
upload stops working until restarted


ihorkhavkin:
I was running Logstash v1.5.6 with output plugin (with arguably minor modifications vs core-v1 https://github.com/ihorkhavkin/logstash-output-s3/tree/v1-patched ) and hitting error like this:

```
[...]

PR #80, opened 2016-05-12
Fix worker dying on upload exception and shutdown

ihorkhavkin:
Fixes issue with upload stopping when `move_file_to_bucket` fails and cleans up shutdown a bit (as case doesn't seem to work on class comparison - http://stackoverflow.com/a/5694333/5657104 )

### DIFF
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -358,7 +358,7 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   private
   def shutdown_upload_workers
     @logger.debug("S3: Gracefully shutdown the upload workers")
-    @upload_queue << LogStash::ShutdownEvent
+    @upload_queue << LogStash::SHUTDOWN
   end
 
   private
@@ -404,7 +404,12 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
         while true do
           @logger.debug("S3: upload worker is waiting for a new file to upload.", :worker_id => worker_id)
 
-          upload_worker
+          begin
+            upload_worker
+          rescue Exception => ex
+            @logger.error('upload_worker unhandled exception', :ex => ex, :backtrace => ex.backtrace)
+            raise LogStash::Error, 'S3: uploader thread exited unexpectedly'
+          end
         end
       end
     end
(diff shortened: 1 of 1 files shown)


ihorkhavkin:
https://github.com/logstash-plugins/logstash-output-s3/issues/79
Will sign CLA after my employer's approval.

ihorkhavkin:
CLA was signed, please let me know if something else needs to be done for check ^ to pass

andrewvc:
@ihorkhavkin thanks for the PR. Apologies for the extreme delay in looking at this, I'm taking a look now :)

andrewvc:
This mostly looks good to merge to me, but I want to hold off for a minute. It looks to me like nothing properly terminates the worker threads on shutdown. That seems problematic. 

The tests all pass but I want to do a manual test to see how that goes. I'm not sure if this is the PR to fix this in (probably not) but I want to think about it over the weekend.

lib/logstash/outputs/s3.rb
```diff
@@ -404,23 +404,36 @@ def configure_upload_workers
...
-        @upload_queue.enq(LogStash::ShutdownEvent)
+        @upload_queue.enq(LogStash::SHUTDOWN)
       else
         @logger.debug("S3: upload working is uploading a new file", :filename => File.basename(file))
         move_file_to_bucket(file)
+      end
+    rescue Exception => ex
+      @logger.error("failed to upload, will re-enqueue #{file} for upload",
+                    :ex => ex, :backtrace => ex.backtrace)
+      unless file.nil?
```
ph: The file cannot be nil, since we are dequeuing stuff.

lib/logstash/outputs/s3.rb
```diff
@@ -404,23 +404,36 @@ def configure_upload_workers
...
   private
   def upload_worker
-    file = @upload_queue.deq
+    file = nil
+    begin
+      file = @upload_queue.deq
 
-    case file
-      when LogStash::ShutdownEvent
+      if file.is_a? LogStash::ShutdownEvent
```
ph: We should do the compare with the `LogStash::SHUTDOWN`

lib/logstash/outputs/s3.rb
```diff
@@ -404,23 +404,36 @@ def configure_upload_workers
         while true do
           @logger.debug("S3: upload worker is waiting for a new file to upload.", :worker_id => worker_id)
 
-          upload_worker
+          begin
+            upload_worker
+          rescue Exception => ex
+            @logger.error('upload_worker unhandled exception', :ex => ex, :backtrace => ex.backtrace)
+            raise LogStash::Error, 'S3: uploader thread exited unexpectedly'
```
ph: I don't think this line solve the problem of the thread dying when an exception is raised inside a thread.
I believe we should only catch the exception in the `#upload_worker` method and not in two different places.

lib/logstash/outputs/s3.rb
ph: Actually, It could be nil when queue `#deq` get interrupted at shutdown.

ph: After discussing with @andrewvc we don't think this case happen in the JVM.

andrewvc:
Merged in #90

PR #81, opened 2016-05-18
Add date_prefix configuration option

omeganon:
Thanks for contributing to Logstash! If you haven't already signed our CLA, here's a handy link: https://www.elastic.co/contributor-agreement/

This change adds an optional new 'date_prefix' configuration option that
will allow for the creation of a date-based prefix in the S3 path. This
will work in combination with the existing 'prefix' option if desired.
Any value for this option is expected to be a time definition suitable
for passing through strftime (e.g '%Y-%m-%d/'). The value of this option
will be inserted between 'prefix' and the filename generated by
logstash. This is particularly useful if you have large volumes of
logs that you wish to keep for long periods of time (archival
purposes). For example, a 5 minute time_file setting will result in
156,960 files for an 18 month retention period for each host sending
data. It is more convenient to work with that number of files,
particularly in GUI based tools, if they are distributed into
smaller groups of files. This will also allow for more granular
loading of files into Elasticsearch.

### DIFF
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -118,6 +118,10 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   # Specify a prefix to the uploaded filename, this can simulate directories on S3.  Prefix does not require leading slash.
   config :prefix, :validate => :string, :default => ''
 
+  # Specify a date-based prefix to the uploaded filename, this can simulate directories on S3.  Prefix does not require leading slash.
+  # The value will be parsed through strftime()
+  config :date_prefix, :validate => :string, :default => ''
+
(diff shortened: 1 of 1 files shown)


joshuaspence:
+1

barqshasbite:
The `prefix` setting fully supports logstash interpolation now, which can accomplish what this PR is trying to do. So, this PR can be closed.

kares:
> The prefix setting fully supports logstash interpolation now, which can accomplish what this PR is trying to do. So, this PR can be closed.

thanks for the PR, ^^ makes sense - we do not want too many prefixes (would be confusing) ... 
ideally one should be enough, if anyone feels prefix can not accomplish this let us know.

PR #85, opened 2016-06-20
removed the delete_on_bucket function entirely

Lswith:
I've signed the CLA and need this feature.

This is not a necessary function and can be safely removed.


karmi:
Hi @Lswith, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-output-s3/pull/85.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

Lswith:
sure

joshuaspence:
+1

ph:
LGTM

PR #86, opened 2016-06-29
Update doc no superusers

ph:
There is no need to have superuser access to run this plugin
Logstash only need to have permission to write to the temporary
directory and access to the file he need to parse.


webmat:
PR is no longer relevant, as this requirement has been removed elsewhere.

PR #87, opened 2016-07-07
Add support for S3 bucket_owner_full_control ACL

dvdmelamed:
The `bucket_owner_full_control` ACL was missing in the supported S3 ACLs though it is supported by the Ruby AWS SDK.

It is useful for cases like running logstash in an EC2 instance that is not the same than the destination account holding the S3 output bucket.


ph:
LGTM

PR #90, opened 2016-08-08
Fix close behavior

andrewvc:
Before workers did not cleanly shutdown resulting in a potential thread leak.

Also includes the patches in #80 , it makes more sense to merge these together as this patch includes spces.

### DIFF
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -135,7 +135,7 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
 
   # Exposed attributes for testing purpose.
   attr_accessor :tempfile
-  attr_reader :page_counter
+  attr_reader :page_counter, :upload_workers
   attr_reader :s3
 
   def aws_s3_config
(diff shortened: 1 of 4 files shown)


lib/logstash/outputs/s3.rb
```diff
@@ -401,26 +401,43 @@ def configure_upload_workers
...
+        while continue do
           @logger.debug("S3: upload worker is waiting for a new file to upload.", :worker_id => worker_id)
 
-          upload_worker
+          begin
+            continue = upload_worker
+          rescue Exception => ex
+            @logger.error('upload_worker unhandled exception', :ex => ex, :backtrace => ex.backtrace)
+            raise LogStash::Error, 'S3: uploader thread exited unexpectedly'
+          end
```
ph: In the `#upload_worker` method we are dealing with the exception, I don't think we need to handle them in this one too?

spec/outputs/s3_spec.rb
```diff
@@ -312,6 +312,33 @@
...
+      end
+
+      it "should remove all worker threads" do
+        subject.do_close
+        sleep 1
+        expect(subject.upload_workers.map(&:thread).any?(&:alive?)).to be false
+      end
+    end
+
     it "doesn't skip events if using the time_file option", :tag => :slow do
```
ph: +1 for the added tests.

lib/logstash/outputs/s3.rb
andrewvc: Good catch! I didn't realize that stud task worked this way :) Will remove this block.

lib/logstash/outputs/s3.rb
```diff
@@ -401,26 +401,38 @@ def configure_upload_workers
...
   private
   def upload_worker
-    file = @upload_queue.deq
+    file = nil
+    begin
+      file = @upload_queue.deq
 
-    case file
-      when LogStash::ShutdownEvent
+      if file.is_a? LogStash::ShutdownEvent
```
ph: Oops, I missed that in the first review, I think we should compare with the value of `LogStash::SHUTDOWN` instead of the class type.

ph:
Small comment, address it and LGTM to merge.

Make sure we rebase off master since there is conflict in this PR.

PR #91, opened 2016-08-19
Fix config value for KMS encryption

wilkystyle:
The existing implementation is not correct for V2 of the AWS SDK (looks like it's using the API from V1).

Updating the plugin to use a string instead of a boolean, which conforms to the SDK docs:
http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Object.html#server_side_encryption-instance_method


karmi:
Hi @wilkystyle, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-output-s3/pull/91.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #96, opened 2016-09-06
new concurrency model :single

ph:
This plugin in the current state is not thread safe and we should rely to the :single delegator.

### DIFF
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -79,7 +79,7 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   config_name "s3"
   default :codec, 'line'
 
-  concurrency :shared
+  concurrency :single
 
   # S3 bucket
   config :bucket, :validate => :string
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -258,6 +258,15 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
 
     begin
       write_on_bucket(test_filename)
+
+      begin
+        remote_filename = "#{@prefix}#{File.basename(test_filename)}"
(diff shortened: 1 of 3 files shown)


lib/logstash/outputs/s3.rb
```diff
@@ -175,6 +177,7 @@ def write_on_bucket(file)
         # prepare for write the file
         object = bucket.objects[remote_filename]
         object.write(fileIO,
+                     :upload_id => remote_filename,
```
ph: AWS, changes the write a bit on patch level.

andrewvc: Should there be a change in the gemspec to account for this?

ph:
I've let the plugin for a few time and there is threadsafe issue in the current code that prevent it to use `shared` so I've taken the path of least resistance and used `single`. The problem is when elements get deleted from disk but are already in the queue. The page counter increment isnt threadsafe too.

The rewrite I did is threadsafe and uses the `shared` model. (see branch https://github.com/ph/logstash-output-s3/tree/feature/cleaner-code)

lib/logstash/outputs/s3.rb
```diff
@@ -79,6 +79,8 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   config_name "s3"
   default :codec, 'line'
 
+  concurrency :single
```
jsvd: humm the PR name is "new concurrency model :shared" and the description says the plugin threadsafe, however this sets the model to `:single`? this means that 1 instance is shared among all threads and access to it is globally synchronized.

ph: yewps, this plugin is in fact in this current state non threadsafe, I've made the plugin run for a few hours to discover that problem.

ph: I will update the description

jsvd:
LGTM

PR #99, opened 2016-09-07
Use correct identifier for Apache License 2.0

andrewgaul:
This addresses a warning during "gem build":

```
WARNING:  WARNING: license value 'Apache License 2.0' is invalid.  Use a
license identifier from http://spdx.org/licenses or 'Nonstandard' for a
nonstandard license.
```


ph:
LGTM, test failling because of this PR https://github.com/logstash-plugins/logstash-output-s3/pull/96

Issue #4, opened 2014-12-22
Dynamic Prefix: Support configuring uploaded file paths/names within a bucket


adamvduke:
The current version of the plugin stores all of the uploaded files in the root of the configured bucket. It would be valuable to be able to configure the path and filename within the bucket to allow for ease of longtime archiving without having to scroll through 1000's of files when viewing the contents of the bucket. Maybe a static configuration, or a proc that can be eval'd before uploading the new file?

```
s3 {
  ...
  path => "{year}/{month}/{day}/{hour}/{host}{uuid}"
  path_proc => '{|time|  "#{time.year}/#{time.month}/#{time.day}/#{time.hour}/#{Socekt.gethostname}#{uuid}'
}
```

pruthvintss:
Hey, is there an update on how this can be done? would like to configure the path with event fields

bradleyphipps:
This would definitely be useful. Outputting to folders within in a bucket based on their type would make data archival/retention/deletion processes easier to manage/safer. Sadly the following config doesn't work: `bucket => "logstash-output-bucket/%{type}"`

radupantiru:
+1

shangliuyan:
+1

wiibaa:
S3 newbie question,
does this request means that 
1. the @prefix config should be able to use event data like `%{type}/` or `%{+YYYY}/%{+MM}/%{dd}/ ??to specify a "folder structure"
2. the filename itself currently generated in [this method](https://github.com/logstash-plugins/logstash-output-s3/blob/master/lib/logstash/outputs/s3.rb#L306-L315) should be able to use event data ?

If this is right, I could dig in further, most probably looking for 1. first
Thoughts ?

adamvduke:
It looks like there are several implementations of this feature to varying degrees. I'm currently not focused on this issue, but perhaps someone on this thread could contribute to one of the open PR's to one/more/all of them along.

https://github.com/logstash-plugins/logstash-output-s3/pull/44
https://github.com/logstash-plugins/logstash-output-s3/pull/59
https://github.com/logstash-plugins/logstash-output-s3/pull/70
https://github.com/logstash-plugins/logstash-output-s3/pull/81

FlorinAndrei:
The `prefix` parameter for the S3 output plugin now (v5.2.0 or newer) supports string interpolation based on field values.

```
s3 {
  prefix => "logstash-test/%{event}/"
}
```

Issue #50, opened 2015-11-11
Permissions on subfolders


jelmstrom:
We are trying to use one bucket for multiple applications, but separating them into subfolders. In AWS it is possible to give permissions to subfolders only, but the plugin requires access rights on the root level of the bucket as well. I guess thats because there are some calls where the prefix is not included. It would make sense to always add the prefix if there is one specified to avoid access to the root level.

Figuring this out was a bit of a trial-and-error exercise, and its not really specified anywhere in the documentation or made obvious in the code.

It would be quite helpful to specify the access rights the plugin needs and that it needs it at root level.

ph:
The problem here is due to the fact that we try to write a test file at the root of the bucket to validate the credentials when we start logstash.

Bad credentials are really the major problem we encounter when we debug stuff with users, I think we could allow people to disable the check when they have special permission on their bucket.

MrPrimate:
For others: You can add the resource ` "arn:aws:s3:::{bucket_name}/logstash-programmatic-access-test-object-*" ` to your policy as this is the path that it tries to write to, rather than granting full bucket access.

cnelson:
I think all it would take to solve this issue is to write the test file to the `prefix` specified in the config rather than the root of the bucket.

Would you merge a PR that would do this?

PR #102, opened 2016-09-16
Refactoring, Dynamic prefix and AWS v2

ph:
**Motivation**
One of the most requested features was adding a way to add dynamic prefixes using the fieldref
syntax for the files on the bucket and also the changes in the pipeline to support shared delegator.
The S3 output by nature was always a single threaded writes but had multiples workers to process the upload, the code was threadsafe when used in the concurrency `:single` mode.

This PR addresses a few problems and provide shorter and more structured code:
- This Plugin now uses the V2 version of the SDK, this make sure we receive the latest updates and changes.
- We now uses S3's `upload_file` instead of reading chunks, this method is more efficient and will uses the multipart with threads if the files is too big.
- You can now use the `fieldref` syntax in the prefix to dynamically changes the target with the events it receives.
- The Upload queue is now a bounded list, this options is necessary to allow back pressure to be communicated back to the pipeline but its configurable by the user.
- If the queue is full the plugin will start the upload in the current thread.
- The plugin now threadsafe and support the concurrency model `shared`
- The rotation strategy can be selected, the recommended is `size_and_time` that will check for both the configured limits (`size` and `time` are also available)
- The `restore` option will now use a separate threadpool with an unbounded queue
- The `restore` option will not block the launch of logstash and will uses less resources than the real time path
- The plugin now uses `multi_receive_encode`, this will optimize the writes to the files
- rotate operation are now batched to reduce the number of IO calls.
- Empty file will not be uploaded by any rotation rotation strategy
- We now use Concurrent-Ruby for the implementation of the java executor
- If you have finer grain permission on prefixes or want faster boot, you can disable the credentials check with `validate_credentials_on_root_bucket`
- The credentials check will no longer fails if we can't delete the file
- We now have a full suite of integration test for all the defined rotation

Fixes: #4 #81 #44 #59 #50


tkampour:
How likely is this to be merged soon?

ph:
@tkampour Since this is a big change, in fact a complete rewrite, it need to go through the review process like any other PRs. Since you have commented on this issue you will get notified when it happen, since its this one of most requested feature I am also eager to get this merged in.

lremurphy:
Awesome work!  Can't wait to try this!

vistorve:
I too am very excited for this change, so excited I am attempting to create a local build of it to get working but I keep running into the following:

Couldn't find any output plugin named 's3'. Are you sure this is correct? Trying to load the s3 output plugin resulted in this error: no such file to load -- concurrent/map", :level=>:error, :file=>"logstash/agent.rb", :line=>"448", :method=>"create_pipeline"}
{:timestamp=>"2016-09-20T16:29:51.292000+0000", :message=>"starting agent", :level=>:info, :file=>"logstash/agent.rb", :line=>"213", :method=>"execute"}

I built the gem and am using that with --no-verify to install. Logstash version=2.4, I am doing this inside a docker if that makes a difference.

suyograo:
@vistorve how did you build this gem?  are you following the steps here? https://github.com/logstash-plugins/logstash-output-s3#2-running-your-unpublished-plugin-in-logstash

vistorve:
I did the steps in 2.2

ph:
@vistorve Nice finding, this is a bug in this PR. I am using **concurrent-ruby**'s map implementation and its available in 1.0.0 >= Logstash 5.0/master ships with it but 2.4 ships with 0.9.2. The dependency is not strict enough in the gemspec this is why you were able to install it but failed to run.

logstash-core depends on a strict version of it in https://github.com/elastic/logstash/blob/2.4/Gemfile.jruby-1.9.lock#L162 so if the plugin requires a newer version the core wont permit it.

@suyograo If we want to make this change works in both 2.4 and 5.0 we have the following solutions:

1. When running in 2.4 Using `threadsafe::hash` that already shipped with logstash, otherwise we use concurrent-ruby's map.

2. Replace concurrent-ruby's map with https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentMap.html which is part of the standard library.

3. release a 2.4.1 with a concurrent-ruby updated.

The other features I used from concurrent-ruby ScheduledTask and the threadpool are available in 0.9.2.

suyograo:
@ph can we do #2 since it looks like a safe option across 2.4 and 5.0?

ph:
@suyograo #2? We have the `temporary_directory` option for quite a while now?

suyograo:
@ph I meant point 2:

> Replace concurrent-ruby's map with https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentMap.html which is part of the standard library.

andrewvc:
FWIW I generally prefer the java util concurrent stuff to the concurrent ruby objects. IMHO the main point of concurrent ruby is to be portable across ruby implementations' concurrency models, which is not a concern for Logstash.

lib/logstash/outputs/s3/size_and_time_rotation_policy.rb
```diff
@@ -0,0 +1,24 @@
...
+        def initialize(file_size, time_file)
+          @size_strategy = SizeRotationPolicy.new(file_size)
+          @time_strategy = TimeRotationPolicy.new(time_file)
+        end
+
+        def rotate?(file)
+          @size_strategy.rotate?(file) || @time_strategy.rotate?(file)
+        end 
+
+        def need_periodic?
```
andrewvc: s/need_periodic/needs_periodic/

lib/logstash/outputs/s3/temporary_file.rb
```diff
@@ -0,0 +1,45 @@
...
+module LogStash
+  module Outputs
+    class S3
+      # Wrap the actual file descriptor into an utility classe
+      # It make it more OOP and easier to reason with the paths.
+      class TemporaryFile
+        extend Forwardable
+        DELEGATES_METHODS = [:path, :write, :close, :size, :fsync]
+
+        def_delegators :@fd, *DELEGATES_METHODS
```
andrewvc: AFAICT we only use a constant here to test if these methods are defined in a later spec. I would just define them inline and remove the spec. I don't think there's a ton of value in specs that check that methods merely exist since at that point we're testing that `Forwardable` isn't broken. 

I don't think we need to do any testing here FWIW and we can simplify this.

lib/logstash/outputs/s3/temporary_file_factory.rb
```diff
@@ -0,0 +1,90 @@
...
+        def increment_counter
+          @counter += 1
+        end
+
+        def current_time
+          Time.now.strftime(STRFTIME)
+        end
+
+        def generate_name
+          filename = "ls.s3.#{Socket.gethostname}.#{current_time}"
```
andrewvc: We need to document that there is a (remote) possibility of collisions between LS instances if this scheme is used. If users use multiple logstashes they should probably give them unique prefixes.

Alternatively we could inject some randomness into here as well.

lib/logstash/outputs/s3.rb
```diff
@@ -63,37 +65,54 @@
...
 
   # S3 bucket
-  config :bucket, :validate => :string
+  config :bucket, :validate => :string, :required => true
 
   # Set the size of file in bytes, this means that files on bucket when have dimension > file_size, they are stored in two or more file.
   # If you have tags then it will generate a specific size file for every tags
   ##NOTE: define size of file is the better thing, because generate a local temporary file on disk and then put it in bucket.
-  config :size_file, :validate => :number, :default => 0
+  config :size_file, :validate => :number, :default => 1024 * 1024 * 5
```
ph: +1, I wanted to change theses names for quite a while, I will keep the other working and mark them as deprecated.

lib/logstash/outputs/s3/temporary_file_factory.rb
```diff
@@ -0,0 +1,90 @@
...
+      # Since the file can contains dynamic part, we have to handle a more local structure to
+      # allow a nice recovery from a crash.
+      #
+      # The local structure will look like this.
+      #
+      # <TEMPORARY_PATH>/<UUID>/<prefix>/ls.s3.localhost.%Y-%m-%dT%H.%m.tag_es_fb.part1.txt.gz 
+      #
+      # Since the UUID should be fairly unique I can destroy the whole path when an upload is complete.
+      # I do not have to mess around to check if the other directory have file in it before destroying them.
+      class TemporaryFileFactory
```
ph: I agree it feel more like a proxy to me, in this case.

lib/logstash/outputs/s3/temporary_file_factory.rb
ph: what do are you suggesting for randomness? 

We have:
- the host
- a date

SecureRandom.uuid?

lib/logstash/outputs/s3/time_rotation_policy.rb
```diff
@@ -0,0 +1,26 @@
...
+        def initialize(time_file)
+          if time_file <= 0
+            raise LogStash::ConfigurationError, "`time_file` need to be greather than 0"
+          end
+
+          @time_file = time_file
+        end
+
+        def rotate?(file)
+          file.size > 0 && Time.now - file.ctime >= time_file
```
s4nch3z: Docs says ```time_file``` is in MINUTES, but it's not like that here

nevins-b:
There hasn't seemed to be any activity on this pull request in a while so I made a [fork](/nevins-b/logstash-output-s3/tree/feature/cleaner-code) and modified the code to address most of the issues raised with this review

ph:
@nevins Thanks can you make a PR with your changes? I will help moving it forward :)

nevins-b:
@ph thanks! opened #105

ph:
I seem to run into a weird jruby thread/autoload issue when running the tests. To be clear the bug was there before but the new logger infra expose it.


https://github.com/jruby/jruby/issues/3920

ph:
@andrewvc This PR has been updated with your requested changes:

- Fix an issue with gzip encoding, its wasn't working, I have added 2 integration test to make sure the behavior is fine.
- Added the code from @nevins-b 
- making sure all the unit test/integration test pass

I believe its ready for the final review.

andrewvc:
Got this error testing with a simple config:

```
[2016-12-07T13:35:33,968][ERROR][logstash.outputs.s3      ] Uploading failed, retrying {:exception=>#<Aws::S3::Errors::MalformedXML: The XML you provided was not well-formed or did not validate against our published schema>, :path=>"/var/folders/x1/v84r4j5j4kx3yvyqrlnhst300000gn/T/logstash/ls.s3.andrewvc-elastic.2016-12-07T12.32.part0.txt", :backtrace=>["/Users/andrewvc/Downloads/logstash-5.0.2/vendor/bundle/jruby/1.9/gems/aws-sdk-core-2.3.22/lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call'", "/Users/andrewvc/Downloads/logstash-5.0.2/vendor/bundle/jruby/1.9/gems/aws-sdk-core-2.3.22/lib/aws-sdk-core/plugins/s3_sse_cpk.rb:19:in `call'", "/Users/andrewvc/Downloads/logstash-5.0.2/vendor/bundle/jruby/1.9/gems/aws-sdk-core-2.3.22/lib/aws-sdk-core/plugins/s3_accelerate.rb:33:in `call'", "/Users/andrewvc/Downloads/logstash-5.0.2/vendor/bundle/jruby/1.9/gems/aws-sdk-core-2.3.22/lib/aws-sdk-core/plugins/param_converter.rb:20:in `call'", "/Users/andrewvc/Downloads/logstash-5.0.2/vendor/bundle/jruby/1.9/gems/aws-sdk-core-2.3.22/lib/seahorse/client/plugins/response_target.rb:21:in `call'", "/Users/andrewvc/Downloads/logstash-5.0.2/vendor/bundle/jruby/1.9/gems/aws-sdk-core-2.3.22/lib/seahorse/client/request.rb:70:in `send_request'", "/Users/andrewvc/Downloads/logstash-5.0.2/vendor/bundle/jruby/1.9/gems/aws-sdk-core-2.3.22/lib/seahorse/client/base.rb:207:in `put_object'", "/Users/andrewvc/Downloads/logstash-5.0.2/vendor/bundle/jruby/1.9/gems/aws-sdk-resources-2.3.22/lib/aws-sdk-resources/services/s3/file_uploader.rb:42:in `put_object'", "/Users/andrewvc/Downloads/logstash-5.0.2/vendor/bundle/jruby/1.9/gems/aws-sdk-resources-2.3.22/lib/aws-sdk-resources/services/s3/file_uploader.rb:49:in `open_file'", "/Users/andrewvc/Downloads/logstash-5.0.2/vendor/bundle/jruby/1.9/gems/aws-sdk-resources-2.3.22/lib/aws-sdk-resources/services/s3/file_uploader.rb:41:in `put_object'", "/Users/andrewvc/Downloads/logstash-5.0.2/vendor/bundle/jruby/1.9/gems/aws-sdk-resources-2.3.22/lib/aws-sdk-resources/services/s3/file_uploader.rb:34:in `upload'", "/Users/andrewvc/Downloads/logstash-5.0.2/vendor/bundle/jruby/1.9/gems/aws-sdk-resources-2.3.22/lib/aws-sdk-resources/services/s3/object.rb:251:in `upload_file'", "/Users/andrewvc/projects/lsp/logstash-output-s3/lib/logstash/outputs/s3/uploader.rb:38:in `upload'", "/Users/andrewvc/projects/lsp/logstash-output-s3/lib/logstash/outputs/s3/uploader.rb:29:in `upload_async'", "org/jruby/RubyProc.java:281:in `call'", "/Users/andrewvc/Downloads/logstash-5.0.2/vendor/bundle/jruby/1.9/gems/concurrent-ruby-1.0.0-java/lib/concurrent/executor/java_executor_service.rb:94:in `run'", "Concurrent$$JavaExecutorService$$Job_1607972051.gen:13:in `run'"]}
```

Config:

```

input {
  generator  { count => 100}
}

filter {
}

output {
  s3 {
    access_key_id => "REDACTED"
    secret_access_key => "REDACTED"
    bucket => "REDACTED"
    size_file => 1048576 # 1MB
    region => "us-east-1"
    canned_acl => public_read
  }

  stdout {
    codec => dots
  }
}
```

ph:
@andrewvc the isses seems to be resolved now, I haven't been able to reproduce it on my side.
We have zoomed on this to dig deeper in the aws SDK framework. Maybe it could have been a temporary hiccups we don't know for sure at this time.

andrewvc:
@ph have you confirmed that the 'gzip' encoding works correctly? When I enable it I don't get gzipped content on S3.

ph:
@andrewvc I've tried it this morning with gzip encoding with the included config, what is happening to you? No files at all? Do you use a fixed amount of events? If you are using the 1mb `size_file` limit it can take a while to generate a gzip file.

If I take a 1mb gzip file and uncompress it, I get a 89 mb text file that contains 2132597 events.

```
input { generator {} }
filter {}
output {
    s3 {
        access_key_id => "redacted"
            secret_access_key => "redacted"
            bucket => "redacte"
            region => "us-east-1"
            encoding => "gzip"
            size_file => 1048576
            canned_acl => public_read
    }
    stdout {
        codec => dots
    }
}
```

andrewvc:
@ph the problem was that the files were uploaded plain encoded :(

andrewvc:
@ph my mistake, it was working right.

lib/logstash/outputs/s3/temporary_file_factory.rb
```diff
@@ -0,0 +1,122 @@
...
+                 IOWrappedGzip.new(::File.open(::File.join(path, key), FILE_MODE))
+               else
+                 ::File.open(::File.join(path, key), FILE_MODE)
+               end
+
+          TemporaryFile.new(key, io, path)
+        end
+
+        class IOWrappedGzip
+          extend Forwardable
```
andrewvc: Can you `require 'forwardable'` at the top of this file?

andrewvc:
Left a few more comments, it's almost good to go!

ph:
@andrewvc I have addressed your latest comment.

ph:
rekicking the travis job

andrewvc:
@ph approved! Great work :)

PR #106, opened 2016-11-22
Support cn-north-1 s3_endpoint

xtao:
Fix s3_endpoint for cn-north-1

### DIFF
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -159,7 +159,7 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
 
   def aws_service_endpoint(region)
     return {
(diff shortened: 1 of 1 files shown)


andrewvc:
@xtao , can you sign our CLA http://www.elasticsearch.org/contributor-agreement/ ? We can't accept any patches until that is done :)

Thanks for the patch, this looks like a worthwhile enhancement!

kares:
still a legit issue, just checked and `cn-northwest-1` also works this way (using a `.cn` suffix)

kares:
issue is defunct (unnecessary - cn regions are supported), for reference see https://github.com/logstash-plugins/logstash-output-s3/pull/108#issuecomment-582339375

PR #108, opened 2016-12-21
fix connection error with cn-north-1 region

### LINKED ISSUE #72: logstash-output-s3 plugin error when region is "cn-north-1"
I think this is a bug.
When I start the logstash, error log: 
"The error reported is: \n initialize: name or service not known"
my logstash config file:

...
region => "cn-north-1"
...

I check the request url: 
https://xxx.s3-cn-north-1.amazonaws.com
But when region is "cn-north-1", the url should be:
https://xxx.s3.cn-north-1.amazonaws.com.cn

Then I check the code, and found: 
vendor/bundle/jruby/1.9/gems/logstash-output-s3-2.0.4/lib/logstash/outputs/s3.rb
ln 143:
:s3_endpoint => region_to_use == 'us-east-1' ? 's3.amazonaws.com' : "s3-#{region_to_use}.amazonaws.com

Then I modify this code, start success.

alexzhangs:
Fixes #72

### DIFF
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -323,7 +323,8 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   end
 
   def aws_service_endpoint(region)
-    { :s3_endpoint => region == 'us-east-1' ? 's3.amazonaws.com' : "s3-#{region}.amazonaws.com"}
+    { :s3_endpoint => region == 'us-east-1' ? 's3.amazonaws.com' : (
+        'cn-north-1' ? "s3.#{region}.amazonaws.com.cn" : "s3-#{region}.amazonaws.com") }
   end
 
   def rotate_if_needed(prefixes)


alexzhangs:
The CLA is signed.

IamGabrielWu:
Hey it looks like your change doesn't consider cn-northwest-1.

alexzhangs:
@IamGabrielWu yep, back then, cn-northwest-1 did not exist.
you may add it to the plugin.

kares:
the setting confused me, but I have confirmed these are unnecessary now.
back the <= 3.2 days, V1 of AWS was pulled in from the mixin but has changed to [V2 since 4.0](https://github.com/logstash-plugins/logstash-output-s3/blob/v4.0.0/lib/logstash/outputs/s3.rb#L90)

V2 version seems to be checking but does not call the `aws_service_endpoint` [hook](https://github.com/logstash-plugins/logstash-mixin-aws/blob/v4.3.0/lib/logstash/plugin_mixins/aws_config/v2.rb#L31)
(`respond_to?` will returns false for a private method)

setting the region simply works using the aws gems : 
```
>> client = Aws::S3::Client.new :region=>"us-east-1"
=> #<Aws::S3::Client>
>> client.config.endpoint
=> #<URI::HTTPS https://s3.us-east-1.amazonaws.com>
>> client = Aws::S3::Client.new :region=>"cn-x-1"
=> #<Aws::S3::Client>
>> client.config.endpoint
=> #<URI::HTTPS https://s3.cn-x-1.amazonaws.com.cn>
```

PR #113, opened 2017-01-04
Typo in cannel_acl usage for the upload options

ph:
The code was referring to `@cannel_acl` instead of `@canned_acl` making the configured not correctly applied to the object.

Fix a layout issue in the changelog.


lib/logstash/outputs/s3.rb
```diff
@@ -288,14 +296,6 @@ def aws_service_endpoint(region)
     { :s3_endpoint => region == 'us-east-1' ? 's3.amazonaws.com' : "s3-#{region}.amazonaws.com"}
   end
 
-  def upload_options
```
andrewvc: Why is this diff so large? Shouldn't it only be 1 line?

ph: I've made the method public to unit test it. :(
There was a layout issue in the changelog so I have corrected it.

Issue #118, opened 2017-01-13
Unsupported encryption method


nevins-b:
I'm getting this error when trying to use v4.0.3:
```
Aws::S3::Errors::InvalidArgument: The encryption method specified is not supported
```

logstash version: 2.4.0
os: centos 7.2

config:
```
input {
  beats {
    port => 5044
  }
}

output {
  s3 {
    bucket => "bucket"
    region => "us-west-2"
    access_key_id => "access"
    secret_key_id => "key"
    server_side_encryption => true
  }
}
```

ph:
fixed in #119

Issue #117, opened 2017-01-13
uninitialized constant LogStash::Util::Loggable


nevins-b:
I'm getting this error when trying to use v4.0.3:
```
fetched an invalid config { :reason=>"Couldn't find any output plugin named 's3'. Are you sure this is correct? Trying to load the s3 output plugin resulted in this error: uninitialized constant LogStash::Util::Loggable", :level=>:error}
```

logstash version: 2.4.0
os: centos 7.2

config:
```
input {
  beats {
    port => 5044
  }
}

output {
  s3 {
    bucket => "bucket"
    region => "us-west-2"
    access_key_id => "access"
    secret_key_id => "key"
    server_side_encryption => true
  }
}
```

ph:
Fixed in #119

Issue #104, opened 2016-11-04
server_side_encryption: 'aws:kms'


tlrdstd:
#62 added support for a boolean `server_side_encryption` flag. If it is `true`, the S3 write uses `:aes256` encryption: https://github.com/logstash-plugins/logstash-output-s3/blob/master/lib/logstash/outputs/s3.rb#L181

My application requires S3 writes to be encrypted via `'aws:kms'` instead of `:aes256`. Could the `server_side_encryption` option please be updated to accept both of these arguments supported by the AWS SDK Object API? (http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Object.html#server_side_encryption-instance_method.)

ph:
fixed in #119

vktor18:
hi, if `server_side_encryption` is true, will be used aes256 but what about the key? I didn't see any specific field where to put the key... Am I missing something? thanks

PR #119, opened 2017-01-13
Two fixes and an enhancement

nevins-b:
This PR addresses #118 and #117.

It also adds configurable storage_class option.

### DIFF
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -130,6 +130,8 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   # Specifies wether or not to use S3's AES256 server side encryption. Defaults to false.
   config :server_side_encryption, :validate => :boolean, :default => false
 
+  config :storage_class, :validate => ["STANDARD", "REDUCED_REDUNDANCY", "STANDARD_IA"], :default => "STANDARD"
+
   # Set the directory where logstash will store the tmp files before sending it to S3
   # default to the current OS temporary directory in linux /tmp/logstash
   config :temporary_directory, :validate => :string, :default => File.join(Dir.tmpdir, "logstash")
@@ -266,6 +268,7 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
     {
       :acl => @canned_acl,
       :server_side_encryption => @server_side_encryption ? "AES256" : nil,
+      :storage_class => @storage_class,
       :content_encoding => @encoding == "gzip" ? "gzip" : nil
     }
   end
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -128,7 +128,10 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
          :default => "private"
 
   # Specifies wether or not to use S3's server side encryption. Defaults to no encryption.
-  config :server_side_encryption, :validate => ["AES256", "aws:kms"]
+  config :server_side_encryption, :validate => :boolean, :default => false
+
+  # Specifies what type of encryption to use when SSE is enabled.
+  config :server_side_encryption_algorithm, :validate => ["AES256", "aws:kms"], :default => "AES256"
 
   # The key to use when specified along with server_side_encryption => aws:kms.
   # If server_side_encryption => aws:kms is set but this is not default KMS key is used.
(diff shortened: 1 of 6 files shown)


lib/logstash/outputs/s3.rb
```diff
@@ -130,6 +130,8 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   # Specifies wether or not to use S3's AES256 server side encryption. Defaults to false.
   config :server_side_encryption, :validate => :boolean, :default => false
 
+  config :storage_class, :validate => ["STANDARD", "REDUCED_REDUNDANCY", "STANDARD_IA"], :default => "STANDARD"
```
ph: @nevins-b Can you add documentation concerning this new option, maybe a high level summary and a link to the complete description at http://docs.aws.amazon.com/AmazonS3/latest/dev/storage-class-intro.html

lib/logstash/outputs/s3.rb
```diff
@@ -183,7 +185,7 @@ def register
       raise LogStash::ConfigurationError, "Logstash must have the permissions to write to the temporary directory: #{@temporary_directory}"
     end
 
-    if @validate_credentials_on_root_bucket && !WriteBucketPermissionValidator.valid?(bucket_resource)
+    if @validate_credentials_on_root_bucket && !WriteBucketPermissionValidator.new(@logger).valid?(bucket_resource)
```
ph: I am fine with that, this make the plugin compatible with 2.4.x.

lib/logstash/outputs/s3.rb
```diff
@@ -265,7 +267,8 @@ def normalize_key(prefix_key)
   def upload_options
     {
       :acl => @canned_acl,
-      :server_side_encryption => @server_side_encryption ? :aes256 : nil,
+      :server_side_encryption => @server_side_encryption ? "AES256" : nil,
+      :storage_class => @storage_class,
       :content_encoding => @encoding == "gzip" ? "gzip" : nil
     }
   end
```
ph: Can you add a test for theses 2 new scenarios?

ph:
@nevins-b Can you take a look at https://github.com/logstash-plugins/logstash-output-s3/issues/104 ? I might be a good time to address it in this PR.

 WDYT?

nevins-b:
Comments added, some more tests added as well and support for #104 added.

I wish there was a library which would let us mock but validate the json request that would be make to to S3.

lib/logstash/outputs/s3.rb
kureus: The logstash-core-plugin-api dependency was left at >= 1.60 for 4.0 meaning 2.4.x stacks, such as mine, could install it. That should have been bumped if it was only 5.x onwards compat! There also seems to be no way to target/install a specific version, so stuck with this broken version till either logstash-core-plugin-api bumped up or 2.4.x backwards compat introduced.

kureus:
The logstash-core-plugin-api dependency was left at >= 1.60 for 4.0 release meaning 2.4.x stacks could install it. That should have been bumped if it was only 5.x onwards compat? There also seems to be no way to target/install a specific version of a plugin, so anyone installing on 2.4.x will be stuck with this broken version till either logstash-core-plugin-api bumped up to force them onto a working version or 2.4.x backwards compat introduced.

ph:
@kureus 

> The logstash-core-plugin-api dependency was left at >= 1.60 for 4.0 release meaning 2.4.x stacks could install it. That should have been bumped if it was only 5.x onwards compat?

This was an honest mistake, but with the changes that @nevins-b did, it makes this plugin backward compatible with 2.4.X

> There also seems to be no way to target/install a specific version of a plugin, so anyone installing on 2.4.x will be stuck with this broken version till either logstash-core-plugin-api bumped up to force them onto a working version or 2.4.x backwards compat introduced.

Nothing prevent you to install a previous version of the plugin using the `bin/logstash-install install --version X.X.X logstash-output-s3`

kureus:
@ph Ah, many thanks for the information :-) Any timescale on this merge?

lib/logstash/outputs/s3.rb
```diff
@@ -127,8 +127,19 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   config :canned_acl, :validate => ["private", "public_read", "public_read_write", "authenticated_read"],
          :default => "private"
 
-  # Specifies wether or not to use S3's AES256 server side encryption. Defaults to false.
-  config :server_side_encryption, :validate => :boolean, :default => false
+  # Specifies wether or not to use S3's server side encryption. Defaults to no encryption.
+  config :server_side_encryption, :validate => ["AES256", "aws:kms"]
```
ph: That's the tricky part, we have to make this change backward compatible with older configuration.

I suggest we keep the `server_side_encryption` as a boolean that will toggle the encryption or not. The encryption algorithm can be set in another field, and the default could be `AES256`, I think `server_side_encryption_algorithm` would be an appropriate name for it.

it will be a bit more verbose but it will make the change backward compatible, and we try to provide sane default configuration out of the box.

Also I've double checked the ruby AWS-SDK [doc](https://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Client.html#put_object-instance_method), I believe they accept both AES256 and aes256, so I am fine using the uppercase.
```
server_side_encryption: "AES256", # accepts AES256, aws:kms
```

ph:
@kureus really soon, it's in the last mile of getting merged thanks @nevins-b  :)

lib/logstash/outputs/s3.rb
nevins-b: makes sense to me, changed pushed.

Issue #122, opened 2017-01-18
Temporary files are not deleted


jarro2783:
It looks like temporary files used for uploading are not being deleted. Over time the number of temporary files in `/tmp/logstash` is growing (I've seen 8000 files in one day). I see messages like this
```
{"level":"DEBUG","loggerName":"logstash.outputs.s3","timeMillis":1484697405218,"thread":"S3 output uploader, file: /tmp/logstash/226d7b5d-ff1d-4486-8eb8-5571ebdd7386/ap-southeast-2/2016-10-24/telegraf/ls.s3.5f5af9cb-6e69-42ca-bb44-3a0b7ea31973.2017-01-17T18.44.part1.txt","logEvent":{"message":"Removing temporary file","file":"/tmp/logstash/226d7b5d-ff1d-4486-8eb8-5571ebdd7386/ap-southeast-2/2016-10-24/telegraf/ls.s3.5f5af9cb-6e69-42ca-bb44-3a0b7ea31973.2017-01-17T18.44.part1.txt"}}
```
but that file still exists.

I think this is the actual cause of what we are seeing in #120. @joshuaspence.

Checking the bucket, I think files are being uploaded correctly, and the uploads are keeping up.

- Version: Logstash 5.1.1, logstash-output-s3 4.0.2
- Operating System: Ubuntu 16.04
- Config File
```
output {
  s3 {
    bucket => "__redacted__"
    codec => "json_lines"
    prefix => "ap-southeast-2/%{+yyyy-MM-dd}/%{type}/"
    region => "ap-southeast-2"
    size_file => 104857600
    time_file => 5
    validate_credentials_on_root_bucket => false
  }
}
```
- Steps to Reproduce: Run logstash for longer than the `time_file` rotation period.

jcsorvasi:
I can second this.

logstash version: docker 5.1.1
s3 output plugin version: 4.0.3
plugin config:

```
output {
  s3 {
    bucket => "some_bucket"
    access_key_id => "some_key"
    secret_access_key => "some_secret"
    region => "eu-west-1"
    codec => "json_lines"
    encoding => "none"
    size_file => "16777216"
    time_file => "1"
    upload_workers_count => "8"
    restore => "true"
  }
}
```

[...]

PR #123, opened 2017-01-18
Correctly cleanup file when a S3 upload is succesfully

ph:
The TemporaryFile#delete! method was raising an exception because we
were trying to close a file that was already close, since the
`on_complete` is executed in a thread nothing was present to catch and
log that specific error and the thread would just end up getting killed.

This PR add guards for closing an already close file, also add test
around the stale file uploader checker to make sure only 1 file is
remaining on disk after rotation.

For future debugging we also log any error that can occur in the
uploader callback.

Fixes: #122 #120

### DIFF
--- a/spec/integration/time_based_rotation_with_stale_write_spec.rb
+++ b/spec/integration/time_based_rotation_with_stale_write_spec.rb
@@ -26,7 +26,7 @@ describe "File Time rotation with stale write", :integration => true do
     clean_remote_files(prefix)
     subject.register
     subject.multi_receive_encoded(batch)
-    sleep(1) # the periodic check should have kick int
+    sleep(5) # the periodic check should have kick in
   end
 
   after do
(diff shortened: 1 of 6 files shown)


spec/integration/time_based_rotation_with_stale_write_spec.rb
```diff
@@ -26,7 +26,7 @@
     clean_remote_files(prefix)
     subject.register
     subject.multi_receive_encoded(batch)
-    sleep(1) # the periodic check should have kick int
+    sleep(5) # the periodic check should have kick in
```
ph: give it a bit more time to do his stuff before doing any assert seems to reduce a lot the noise.

joshuaspence:
Should we be using `FileUtils::rm_r` instead of `FileUtils::rm_rf`? It seems that the latter method ignores `StandardError`.

ph:
@joshuaspence This is a good point, I will make another PR with `rm_r`, I would prefer to be notified if something is wrong.

- Wrong permission?
- Trying to delete a non existing file?

joshuaspence:
Thanks!

PR #128, opened 2017-02-17
Use GzipWriter.pos to get file size for gzip encoded files

PedroMDuarte:
This pull request fixes the issue where empty gzip files were being uploaded to S3 when using the gzip encoding.  The root of the problem is that a gzip file has a non-zero size even right after it is opened.  To solve the issue we make use of the `pos()` function of the `GZipWriter` class.  See https://ruby-doc.org/stdlib-1.9.3/libdoc/zlib/rdoc/Zlib/GzipWriter.html. 

I first created a simple test case that exposed the bug, and then verified that my proposed solution takes care of the problem.

### DIFF
--- a/lib/logstash/outputs/s3/temporary_file_factory.rb
+++ b/lib/logstash/outputs/s3/temporary_file_factory.rb
@@ -115,7 +115,7 @@ module LogStash
               0
             else
               @gzip_writer.flush
-              @gzip_writer.to_io.size_file
+              @gzip_writer.to_io.size
             end
           end
 
--- a/lib/logstash/outputs/s3/temporary_file_factory.rb
+++ b/lib/logstash/outputs/s3/temporary_file_factory.rb
@@ -109,7 +109,14 @@ module LogStash
 
           def size
             # to get the current file size
-            @gzip_writer.pos
+            if @gzip_writer.pos == 0
+              # Ensure a zero file size is returned when nothing has
+              # yet been written to the gzip file.
+              0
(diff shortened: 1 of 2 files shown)


karmi:
Hi @PedroMDuarte, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-output-s3/pull/128.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PedroMDuarte:
I signed the CLA after submitting this pull request.  Please let me know if there is anything else I need to do on my end to make sure the CLA signature check is cleared.

PedroMDuarte:
ok, just added my other email to my github profile.

PedroMDuarte:
I just realized that using the solution that I have suggested leads to another bug.  If we rely on `.pos` for the file size,  then rotation by size rotates gzip files based on the amount of input bytes written to the file, and not based on the size of the compressed gzip file.   This could be left as is and clarified in the docs,  or we could make it return the `@gzip_writer.pos` value only when it is zero, and otherwise return `@gzip_writer.to_io.size`

PedroMDuarte:
I have updated the PR so that `@gzip_writer.pos` is used only when nothing has yet been written to the gzip file.   This will allow determining file rotation by size based on the size of the compressed content.

ph:
LGTM

PR #129, opened 2017-02-23
S3 Client Signature Version Attribute Name

chadbean:
It looks like the attribute `s3_signature_version` was part of the AWS Ruby SDK v1. The v2 SDK uses `signature_version` as the attribute name. Essentially this fixes an issue where I was unable to send the logs to a bucket.

> Ruby SDK - Version 2: Set the signature_version parameter to v4 when constructing the client.
> s3 = Aws::S3::Client.new(signature_version: 'v4')

See: http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingAWSSDK.html#specify-signature-version

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -2,6 +2,7 @@ The following is a list of people who have contributed ideas, code, bug
 reports, or in general have helped logstash along its way.
 
 Contributors:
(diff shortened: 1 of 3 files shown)


ph:
LGTM

PR #47, opened 2015-11-03
Remove reference to "format" in example

untergeek:
fixes #46

PR #135, opened 2017-04-12
Fixed Usage

st1t:
I think that it is better to change format option of "Usage" to codec option.


st1t:
Sorry.
I overlooked the correction below.
https://github.com/logstash-plugins/logstash-output-s3/pull/47

Why is format still left in official documentation?
https://www.elastic.co/guide/en/logstash/current/plugins-outputs-s3.html

andrewvc:
Thanks for the PR @st1t ! It looks like we already fixed this https://github.com/logstash-plugins/logstash-output-s3/blob/master/lib/logstash/outputs/s3.rb#L74. Apologies for the delay.

PR #28, opened 2015-08-27
Clarify documentation regarding size_file and time_file

pctj101:
Reworded the comments to make it easier to read / understand.

PR #137, opened 2017-04-18
Replace "format" with "codec" in the example

radu-gheorghe:


jordansissel:
<3 documentation.

I was pretty sure we already fixed this, and we did in #47, but it came back a few months later accidentally in #28.

thanks for fixing this!

radu-gheorghe:
You're welcome, Jordan!

PR #141, opened 2017-04-26
Fixes issues with the `canned_acl` option

ph:
The list of canned acl was not the official one and made the push of
document to S3 fail, this could be considered to be a backward
incompatible changes but if you used something else than private it
would not have worked correctly.

Fix an issue with the `canned_acl` not correctly passed to the crash
uploader when the `restore` option was set to true

Thanks for contributing to Logstash! If you haven't already signed our CLA, here's a handy link: https://www.elastic.co/contributor-agreement/


suyograo:
LGTM

Issue #143, opened 2017-05-16
S3 upload retry infinite loop


marcinc:
It seems that in crash restore scenario plugin can potentially go into an infinite loop when a file it expects no longer exists on disk. It could be potentially a result of some race condition when LS crash happens whereby file gets removed from disk before recovery threadpool gets to process it? In such situation any attempt to upload non-existent file to S3 would fail with exception which in turn logs error and [`retry`](https://github.com/logstash-plugins/logstash-output-s3/blob/7aa5a03baa3a5f74722d3c320d629b5e9e1b257f/lib/logstash/outputs/s3/uploader.rb#L39-L46) immediately. Note: Uploader currently doesn't utilise `TIME_BEFORE_RETRYING_SECONDS` even though it's defined.  

Also I wonder whether [`LogStash::Outputs::S3#upload_file`](https://github.com/logstash-plugins/logstash-output-s3/blob/7aa5a03baa3a5f74722d3c320d629b5e9e1b257f/lib/logstash/outputs/s3.rb#L341-L345) should also delete zero-byte files? `upload_file` only ever gets called on `close` or file rotation. 

- Version: 4.0.7
- Operating System: Fedora 25
- Config File:

```
output {
  s3{
    id => "some_id"
    access_key_id => "aws-key-id"
    secret_access_key => "aws-secret-key"
    bucket => "bucket-name"
    region => "eu-west-1"
    restore => true
    time_file => "10"
    size_file => "10485760"
    canned_acl => "private"
    codec => "json_lines"
    server_side_encryption => true
    server_side_encryption_algorithm => "aws:kms"
    validate_credentials_on_root_bucket => false
  }
}
```

- Sample Log:

```
[...]

PR #144, opened 2017-05-16
Handle file not found exception on S3 upload

marcinc:
This PR addresses https://github.com/logstash-plugins/logstash-output-s3/issues/143:
* Handle uploader missing file exception.
* Wait `TIME_BEFORE_RETRYING_SECONDS` before each file upload retry.

### DIFF
--- a/lib/logstash/outputs/s3/uploader.rb
+++ b/lib/logstash/outputs/s3/uploader.rb
@@ -36,13 +36,16 @@ module LogStash
           begin
             obj = bucket.object(file.key)
             obj.upload_file(file.path, upload_options)
+          rescue Errno::ENOENT => e
+            logger.error("File doesn't exist! Unrecoverable error.", :exception => e.class, :message => e.message, :path => file.path, :backtrace => e.backtrace)
           rescue => e
             # When we get here it usually mean that S3 tried to do some retry by himself (default is 3)
             # When the retry limit is reached or another error happen we will wait and retry.
             #
             # Thread might be stuck here, but I think its better than losing anything
             # its either a transient errors or something bad really happened.
-            logger.error("Uploading failed, retrying", :exception => e.class, :message => e.message, :path => file.path, :backtrace => e.backtrace)
+            logger.error("Uploading failed, retrying.", :exception => e.class, :message => e.message, :path => file.path, :backtrace => e.backtrace)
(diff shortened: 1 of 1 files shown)


ph:
Good find! @marcinc Can you sign the CLA, I will take a look after.

marcinc:
CLA signed. Cheers.

ph:
Code LGTM, I am OK to add the handling of this specific case.


> It could be potentially a result of some race condition when LS crash happens whereby file gets removed from disk before recovery threadpool gets to process it? In such situation any attempt to upload non-existent file to S3 would fail with exception which in turn logs error and retry immediately. Note: Uploader currently doesn't utilise TIME_BEFORE_RETRYING_SECONDS even though it's defined.

I am still trying to find how this case can happen, I don't think we could have a race, the only thing I could see is the file was deleted externally before the process was able to upload it?


Do you have multiple S3 outputs that shares the same *temporary directory*? Or do you have multiple Logstash instances that use the same *temporary directory*?

marcinc:
@ph temp files aren't touched by anything else so scenario where they get deleted externally is very unlikely.  

> Do you have multiple S3 outputs that shares the same temporary directory? Or do you have multiple Logstash instances that use the same temporary directory?

I run multiple s3 outputs (3 in this case) and all of those share the same temp directory.

Could it be that recovery thread pool picks up only after the periodic check happened which can potentially remove those files earlier?

ph:
> I run multiple s3 outputs (3 in this case), and all of those share the same temp directory.

I think the following happen, the three outputs share the same temporary directory, and all of them have the *restore from a crash* option. So all of them read for the temporary directory and try to resume the upload operation. One of them completes the upload of the file before the others, when the other try to upload the document the file will be deleted.

So I think we should merge this PR and create another issue to make sure the temporary directory is locked to one s3 output.


Also depending on your configuration, you might not need to have multiple s3 outputs, maybe you can use a dynamic prefix instead. Assuming you are sending everything to the same bucket.

marcinc:
@ph yeah, I think this makes total sense. I'll try to experiment with that. Will report back on the findings.

marcinc:
@ph your theory proved to be correct. Combined multiple s3 outputs into a single one solved the issue. 

> So I think we should merge this PR and create another issue to make sure the temporary directory is locked to one s3 output.

👍  to the above :)

PR #152, opened 2017-07-20
Add "bucket-owner-full-control" canned ACL

mool:
Add support for the `bucket-owner-full-control` ACL. It's very useful when you are writing files in a bucket owned by a different AWS account.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -64,7 +64,7 @@ output {
      size_file => 2048                        (optional) - Bytes
      time_file => 5                           (optional) - Minutes
      codec => "plain"                         (optional)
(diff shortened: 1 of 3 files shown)


mool:
I've signed the CLA twice, but I still see the the check failing.

aresen:
This looks good, any chance it'll be included soon?

sarneaud:
Thanks for this, but in our case we need `bucket-owner-read`.  Here are all the ACLs available now:
https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl

jsvd:
done in https://github.com/logstash-plugins/logstash-output-s3/pull/158

PR #154, opened 2017-08-21
Document prefix interpolation.

jmcarp:
Document logstash sprintf interpolation for `prefix` and give an example.

cc @sharms


guyboertje:
LGTM

PR #158, opened 2017-09-21
Add more S3 bucket ACLs

### LINKED ISSUE #76: S3 output Plugin does not supports canned_acl => "bucket_owner_full_control"
We need the Outputted file from Logstash to S3 to be fully controlled by bucket owner .

Currently only Below Options are available : 

canned_acl => "private"                  (optional. Options are "private", "public_read", "public_read_write", "authenticated_read". Defaults to "private" )

We should add support for  "bucket_owner_full_control"

MrPrimate:
There are more canned ACL's available, and I would like to use them when storing logs.

Expands #152 Resolves #76

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -64,7 +64,7 @@ output {
      size_file => 2048                        (optional) - Bytes
      time_file => 5                           (optional) - Minutes
      codec => "plain"                         (optional)
-     canned_acl => "private"                  (optional. Options are "private", "public-read", "public-read-write", "authenticated-read". Defaults to "private" )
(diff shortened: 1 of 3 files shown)


randomvariable:
Is there a status update on this?

jordansissel:
@randomvariable what status are you looking for? I don't think anyone's reviewed this, so it waits for that review (and testing, etc).

If you have energy to help PR in its journey, it could be helpful if you test this and report your experience.

bmandrade:
Hey guys, I used logstash s3 output to put objects in a client aws s3 bucket, but I had to use this patch for the client be able to get the objects in it's own bucket.

Besides that, I also needed to add the `PutObjectAcl` action to aws policy, not just `PutObject` as the documentation said.

If you need more info, let me know, to help moving this PR forward.

cc:
@jordansissel 
@MrPrimate  (Thanks a lot for this one 👍 )

jsvd:
Many thanks for the well put PR @MrPrimate

PR #165, opened 2017-11-15
Update description of file_size

MasterCarl:
I found the documentation pretty confusing, so I tried to rewrite it to better reflect that size_file triggers a log rotation. This is more in line with the documentation found at https://github.com/logstash-plugins/logstash-output-s3/blob/master/lib/logstash/outputs/s3.rb

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -252,8 +252,7 @@ specified here
   * Value type is <<number,number>>
   * Default value is `5242880`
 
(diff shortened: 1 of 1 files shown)


MasterCarl:
@karmi For the record, I did sign the CLA

kares:
current docs, already, includes the following line:

> When the size of a local temporary file reaches size_file, a log rotation is triggered and  the current log "part" it pushed to s3. It is then deleted from local temporary storage.

so I am closing this one

PR #185, opened 2018-05-30
Revert rake dependency pin

robbavey:
The rake dependency pinning seems to break upstream tests in 6.x. Reverting


jakelandis:
LGTM

robbavey:
Thanks @jakelandis

PR #192, opened 2018-09-10
Adding possibility to include file name as a part of prefix

akh00:
Sometimes it's necessary to put each log event into separate file in s3 with an exact file name formed from an event fields. The existing "prefix" property can be used for that


treknado:
this feature is highly needed to comply naming conventions in s3. please mergie this.

PR #193, opened 2018-09-11
Fix leak of file handles that prevents gzip files from being cleaned up

yaauie:
Use `Zlib::GzipWriter::new(IO)` instead of `Zlib::GzipWriter::open(Object)`, since
the latter effectively creates an additional file handle when passed an `IO`. This
ensures that when we send `Zlib::GzipWriter#close`, the original `IO` is closed in
turn.

Resolves: https://github.com/logstash-plugins/issues/190


yaauie:
bumped version and added changelog note, will merge/publish once the build goes green

PR #209, opened 2019-07-15
Update doc to update S3 auth requirements

zimbabao:
S3 auth also requires session_token is MFA is enabled. 

Signed https://www.elastic.co/contributor-agreement/

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -31,7 +31,7 @@ option enabled, each output should define its own 'temporary_directory'.
 
 ===== Requirements
 
-* Amazon S3 Bucket and S3 Access Permissions (Typically access_key_id and secret_access_key)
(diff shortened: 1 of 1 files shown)


docs/index.asciidoc
```diff
@@ -31,7 +31,7 @@ option enabled, each output should define its own 'temporary_directory'.
 
 ===== Requirements
 
-* Amazon S3 Bucket and S3 Access Permissions (Typically access_key_id and secret_access_key)
+* Amazon S3 Bucket and S3 Access Permissions (Typically access_key_id, secret_access_key and session_token incase MFA is enabled)
```
robbavey: Maybe we could rephrase this - to something like

```
(Typically `access_key_id` and `secret_access_key`. A `session_token` is also required if Multi Factor Authentication is enabled, or other temporary security credentials are in use that mandate the use of a session token)
```
cc @karenzone

PR #198, opened 2018-11-06
Clarify endpoint documentation

Pectojin:
Clarified the need to explicitly declare HTTP or HTTPS in the endpoint URL.

PR #210, opened 2019-08-20
Bump to v4.1.10 to pick up doc changes

karenzone:
Version bump to pick up doc changes from #198


karenzone:
Published gem
Screen Shot 2019-08-20 at 3 17 08 PM

Issue #134, opened 2017-03-28
Make S3 file name template configurable


drscre:
We are uploading logs to S3. Another consumer periodically collects those file.
Consumer expects that most recently uploaded files appears last in the file list when sorted by name.
So consumer just remembers which file name he downloaded last and queries S3 only for files with filename > last downloaded file.

Newer plugin version use a random UUID in the file name, so the files are no longer name-sortable.

js77usf:
Also affecting our project.  We cannot search files by date.  S3 UI only does a "starts with" on the file name when the search criteria is provided.  If the timestamp portion of the auto-generated file name is placed AFTER the GUID, then there is no way to search for files written to S3 on a certain date.  Older versions of the plugin didn't have this problem.

Additionally, in the S3 UI, when sorting the files by date using the column headers, the sort only affects the data currently displayed on the page.  It does NOT requery for all files, apply the sort, and then update the page with new data.

cgough:
This basically makes the s3 logstash outpush unusable for us -- we need to preserve the filenames we are uploading. Otherwise, very difficult to make sense of output files without a descriptive prefix.

JathinSanghvi:
Atleast there should be an option to have ip instead of guid as done in older versions. This is breaking lots of functionality. Current implementations makes it really tough to retrieve/list recently created files.

lucasbasquerotto:
This issue is opened for about one year and a half, is there any reason it's taking so much to solve it?

If the implementation may take a long time, then at least change the order from `...[UUID].[Time]...` to `...[Time].[UUID]...`, in the meantime.

qingkunl:
> This issue is opened for about one year and a half, is there any reason it's taking so much to solve it?
> 
> If the implementation may take a long time, then at least change the order from `...[UUID].[Time].[Part]...` to `...[Time].[Part].[UUID]...`, in the meantime.

Would love to at least see this change as well.

PR #215, opened 2019-12-18
Added `filename` as a configuration option.

barqshasbite:
This new option can be used to specify how files are named when
uploaded to S3. It supports logstash string interpolation (sprintf)
so can be used to generate unique filenames based on the data from
an event.

Addresses #134

### DIFF
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -151,8 +151,12 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   config :temporary_directory, :validate => :string, :default => File.join(Dir.tmpdir, "logstash")
 
   # Specify a prefix to the uploaded filename, this can simulate directories on S3.  Prefix does not require leading slash.
-  # This option support string interpolation, be warned this can created a lot of temporary local files.
+  # This option supports logstash string interpolation with sprintf. Be warned this can create a lot of temporary local files.
   config :prefix, :validate => :string, :default => ''
+
+  # Specify a filename format to use for uploaded files. If not defined, a unique filename is generated. Invalid characters are replaced with underscores (_).
+  # This option supports logstash string interpolation with sprintf. If you do not configure a unique filename using interpolation, the plugin may overwrite the same file each time an S3 upload takes place.
+  config :filename, :validate => :string, :default => ''
 
   # Specify how many workers to use to upload the files to S3
   config :upload_workers_count, :validate => :number, :default => (Concurrent.processor_count * 0.5).ceil
@@ -197,6 +201,11 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
         raise LogStash::ConfigurationError, "Prefix must not contains: #{PathValidator::INVALID_CHARACTERS}"
       end
     end
+    unless filename.empty?
+      if !PathValidator.valid?(filename)
+        raise LogStash::ConfigurationError, "Filename must not contains: #{PathValidator::INVALID_CHARACTERS}"
+      end
+    end
 
     if !WritableDirectoryValidator.valid?(@temporary_directory)
       raise LogStash::ConfigurationError, "Logstash must have the permissions to write to the temporary directory: #{@temporary_directory}"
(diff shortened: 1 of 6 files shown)


lib/logstash/outputs/s3.rb
```diff
@@ -197,6 +201,11 @@ def register
         raise LogStash::ConfigurationError, "Prefix must not contains: #{PathValidator::INVALID_CHARACTERS}"
       end
     end
+    unless filename.empty?
```
barqshasbite: Should be @filename...

Forbzy:
@barqshasbite I was looking for this functionality. I'm pleased to see someone was working on it. It would be very useful to be able to set file names. Can you say what the current conflict issues are?

BillYoungman:
I noticed that the last commit for this feature failed on December 19, 2019 has anything been attempted since then. From what I'm seeing I'm not the only one who is interested in this feature.

Thanks,
Bill

barqshasbite:
If I remember correctly, the build initially failed for reasons not related to my change (other builds were also failing at the time).

Since then, I have not revisited this change. In its current state, it is working for my limited use case so I did not take it any further. My use-case being uploading single, write-once, JSON files to S3 with a unique name.

I put up the pull request knowing that other people were interested in the feature and may want to use my variation. It is incomplete, though. In that it does not support the `size_file` and `time_file` configuration options for rolling over filenames with a `partN` filename component. It will always upload with the same filename, so could potentially overwrite existing files if you do not have unique naming setup. Adding support for the `partN` filename component would round this out and make it a more complete feature.

BillYoungman:
Our company processes and calculates Sales & Consumer Use Taxes for our clients so on average we get about 10,000 transactions per second that are being moved into Elasticsearch indexes via Filebeat but in addition to this we are moving these transactions in AWS s3 buckets for use in calculating client metric data. The default s3 naming convention of 'ls.s3.xxx' was making it difficult to work with those files hence our need for custom naming so I was really glad when I came across this feature request.

I took your code and after creating a new version of the plug-in in our development environment copied it into that new local version modifying my logstash.yml file to point to this local plug-in. I then made the following modifications to the temporary_file_factory.rb file:

`#     name = filename == "" ? generate_name : filename
	name = filename == "" ? generate_name : generate_custom_name(filename)
`

Created new method:
`        def generate_custom_name(filename)
		  filename = "#{filename}.#{SecureRandom.uuid}.#{current_time}"

          if tags.size > 0
            "#{filename}.tag_#{tags.join('.')}.part#{counter}.#{extension}"
          else
            "#{filename}.part#{counter}.#{extension}"
          end
        end
`
I ran it through extensive testing using JMeter and for us it is working like a charm although I have noticed a couple of things and they are for both the original version and my version of the plug-in neither one are honoring the size_file tag but we think that it is more a case that data is coming in so fast that by the time the rotate / upload is triggered the file size is larger than what is set in the logstash.conf both are honoring the time_file value.

I apologize for the verbosity of this post but I wanted to put some context around what I did and what we saw.

Forbzy:
It good to hear work on this feature is continuing. @BillYoungman is your version of the plugin still up to date with new changes to the Master branch? For my use case I would need the time_file option to work because I'd need to output data hourly and daily. Does this option work for your version @BillYoungman ?

BillYoungman:
@Forbzy it is still in my local file system as this is my first attempt at doing any work like this and wasn't entirely sure of the process involved to officially work in here so didn't want to do anything that might be unauthorized. That being said I do have awhile back I did sign the letter to become a contributing developer.

Let me do some more testing focusing on the time_file property in particular and will post my findings.

BillYoungman:
@barqshasbite not sure if this is the correct place for questions if it isn't please direct me to the right place - thanks.
Adding @Forbzy 

But here goes--
time_size is working but upon closer testing of just the file_size tag (I was using size_and_time in my earlier testing which was actually masking the size variable) it is not working for the call to generate_custom_name(filename) - it never rolls the files over however when I let the plugin use the default naming method call generate_name it's working fine.

When I run my tests with the default method call I see calls to
`		if @rotation.rotate?(temp_file)
			@logger.info("Rotate file",
                       :strategy => @rotation.class.name,
                       :key => temp_file.key,
                       :path => temp_file.path)

			upload_file(temp_file)
			factory.rotate!
        end
`

When I set a custom filename this is not being called at all. So my question is where / how is the default pattern call working but the custom pattern isn't.

Been struggling with this all day and it isn't helping that this is my first foray into Ruby as well.

Thanks,
Bill

Forbzy:
@BillYoungman Is there no way of just allow us to set the filename our selves, for instance you can do this with the file output plugin when setting the path.

yogevyuval:
@Forbzy @BillYoungman @barqshasbite 
This feature is very much needed, the plugin can be useless without it for many use cases as can be seen in the different threads.

Any update about this PR? Is there something that can be done to make this happen? Happy to help if needed

webminster:
I'd like to upvote this as well... I'm trying to use Logz.io with its S3 bucket shipper, and it wants S3 file names to be in ascending sort order.  With the random name, I can't honor that requirement.

yjagdale:
@barqshasbite  - looks like this plugin is no longer maintained. Can you please share a gem file that people can install manually and use?

PR #217, opened 2020-02-05
Refactor: cleanup unused (old) method

kares:
- cleanup unused (old) methods
  was used in V1 - no longer relevant with aws-sdk 2.x (explanation at https://github.com/logstash-plugins/logstash-output-s3/pull/108#issuecomment-582339375)
  (since its private it never got to be called from the mixin)

with https://github.com/logstash-plugins/logstash-mixin-aws/pull/42 this caused some aws gem newbie confusion ...

### DIFF
--- a/spec/outputs/s3_spec.rb
+++ b/spec/outputs/s3_spec.rb
@@ -204,4 +204,27 @@ describe LogStash::Outputs::S3 do
       subject.multi_receive_encoded(events_and_encoded)
     end
   end
+
+  describe "aws service" do
+    context 'us-east-1' do
+      let(:region) { 'us-east-1' }
+      it "sets endpoint" do
+        expect( subject.send(:bucket_resource).client.config.endpoint.to_s ).to eql 'https://s3.us-east-1.amazonaws.com'
+      end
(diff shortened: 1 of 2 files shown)


kares:
please also review: https://github.com/logstash-plugins/logstash-mixin-aws/pull/42 who ever lands here ...

spec/outputs/s3_spec.rb
```diff
@@ -204,4 +204,27 @@
...
+
+  describe "aws service" do
+    context 'us-east-1' do
+      let(:region) { 'us-east-1' }
+      it "sets endpoint" do
+        expect( subject.send(:bucket_resource).client.config.endpoint.to_s ).to eql 'https://s3.us-east-1.amazonaws.com'
+      end
+    end
+
+    context 'us-east-1' do
```
andsel: should it be `ap-east-1` instead of `us-east-1`?

kares:
```suggest
context 'ap-east-1' do
```

andsel:
LGTM

Issue #216, opened 2020-01-21
allow opting out from infinite upload retries


kares:
in some cases user might not want to keep on retry-ing upload forever.
the logic makes sense for "transient" errors (e.g. network/service failure) but is potentially problematic (generating a lot of noise at minimum) for arbitrary raises.

we potentially need to investigate the class of exceptions coming from s3 library.

this feature has been requested previously e.g. https://github.com/logstash-plugins/logstash-output-s3/issues/169 and https://github.com/logstash-plugins/logstash-output-s3/issues/160 ... the later case is specific to `Errno::ENOENT` which is now handled as a non-retry case https://github.com/logstash-plugins/logstash-output-s3/pull/144

Issue #169, opened 2018-02-20
Limit retry interval


Sjaak01:
Hi,

Is it possible to set/limit the retry interval when connecting to s3 fails?

In my case I have a firewall blocking access to s3 over one of my WAN links (low bandwidth, very high costs). However the constant retrying is generating several hundreds MB of DNS traffic an hour.

If a upload fails I don't want a retry for another hour or try uploading again when the next file is due.

idsvandermolen:
duplicate of https://github.com/logstash-plugins/logstash-output-s3/issues/160

PR #218, opened 2020-02-05
Feat: Added retry_count and retry_delay config

kares:
initially wanted to filter out specific errors that make sense to get retried (e.g. only rescue-ing `Aws::S3::Errors::ServiceError` and inspecting `error.code`) but possibilities are wild.

resolves https://github.com/logstash-plugins/logstash-output-s3/issues/216 (also closes https://github.com/logstash-plugins/logstash-output-s3/issues/169)

### DIFF
--- a/lib/logstash/outputs/s3/uploader.rb
+++ b/lib/logstash/outputs/s3/uploader.rb
@@ -6,7 +6,7 @@ module LogStash
   module Outputs
     class S3
       class Uploader
-        TIME_BEFORE_RETRYING_SECONDS = 1
+
         DEFAULT_THREADPOOL = Concurrent::ThreadPoolExecutor.new({
                                                                   :min_threads => 1,
                                                                   :max_threads => 8,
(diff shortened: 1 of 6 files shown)


lib/logstash/outputs/s3/uploader.rb
```diff
@@ -44,15 +46,22 @@ def upload(file, options = {})
             #
             # Thread might be stuck here, but I think its better than losing anything
             # its either a transient errors or something bad really happened.
-            logger.error("Uploading failed, retrying.", :exception => e.class, :message => e.message, :path => file.path, :backtrace => e.backtrace)
-            sleep TIME_BEFORE_RETRYING_SECONDS
-            retry
+            if tries < @retry_count
+              tries += 1
+              logger.warn("Uploading failed, retrying (#{tries}).", :exception => e.class, :message => e.message, :path => file.path, :backtrace => e.backtrace)
```
kares: as retry logging seems to "annoy" users I decreased logger.level to **warn**
(and added a new error message when retrying is done)
but users not configuring `retry_count` will only see logging downgraded to warn level.
if that is a potential problem and we should keep error level here, I am fine to do that ...

jsvd: We default logging to info, so by default it will still show as before, but I don't think it's a big concern as logging levels can be tuned per plugin

Btw, Maybe adding the extra information about total number of retries is also useful. for example:
```suggestion
              logger.warn("Uploading failed, retrying (##{tries} of #{@retry_count})", :exception => e.class, :message => e.message, :path => file.path, :backtrace => e.backtrace)
```

to produce:
```ruby
logger.warn("Uploading failed, retrying (#1 of 4)", :exception => "...", :message => "..", :path => "..", :backtrace => ["..."])
logger.warn("Uploading failed, retrying (#2 of 4)", :exception => "...", :message => "..", :path => "..", :backtrace => ["..."])

```

kares: cool, thanks. we could at least tell users to raise the plugin's level to error :wine_glass:

PR #220, opened 2020-03-05
[Doc] tried to document better the setting

andsel:
Try to improve the description of `time_file` setting

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -385,9 +385,8 @@ default to the current OS temporary directory in linux /tmp/logstash
   * Default value is `15`
 
 Set the time, in MINUTES, to close the current sub_time_section of bucket.
-If you define file_size you have a number of files in consideration of the section and the current tag.
-0 stay all time on listener, beware if you specific 0 and size_file 0, because you will not put the file on bucket,
-for now the only thing this plugin can do is to put the file when logstash restart.
+If <<plugins-{type}s-{plugin}-rotation_strategy,`rotation_strategy`>> is set to `time` or `size_and_time`, then `time_file` cannot be set to 0. 
+Otherwise, the plugin raises a configuration error.
 
 [id="plugins-{type}s-{plugin}-upload_multipart_threshold"]
 ===== `upload_multipart_threshold`
--- a/lib/logstash/outputs/s3.rb
+++ b/lib/logstash/outputs/s3.rb
@@ -114,9 +114,8 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
   config :size_file, :validate => :number, :default => 1024 * 1024 * 5
 
   # Set the time, in MINUTES, to close the current sub_time_section of bucket.
-  # If you define file_size you have a number of files in consideration of the section and the current tag.
-  # 0 stay all time on listerner, beware if you specific 0 and size_file 0, because you will not put the file on bucket,
-  # for now the only thing this plugin can do is to put the file when logstash restart.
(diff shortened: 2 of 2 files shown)


andsel:
Please @karenzone could give an eye if sounds more meaningful and readable, thanks

karenzone:
@andsel Thanks for kicking this off.

At line 376:  
>If you define file_size you have a number of files in consideration of the section and the current tag.

Is this what it means? 
>If you define `size_file`, the section and the current tag also factor into the number of files created.

I know this isn't currently part of the PR, but it's closely related.

docs/index.asciidoc
```diff
@@ -374,8 +374,8 @@ default to the current OS temporary directory in linux /tmp/logstash
 
 Set the time, in MINUTES, to close the current sub_time_section of bucket.
 If you define file_size you have a number of files in consideration of the section and the current tag.
-0 stay all time on listener, beware if you specific 0 and size_file 0, because you will not put the file on bucket,
-for now the only thing this plugin can do is to put the file when logstash restart.
+If you also define file_size you have a number of files related to the section and the current tag.
+If it's valued 0 and `rotation_strategy` is `time` or `size_and_time` then the plugin reaise a configuration error.
```
karenzone: Some settings are not compatible. If `rotation_strategy` is set to `time` or
`size_and_time`, then `time_file` cannot be set to `0`.  Otherwise, the plugin
raises a configuration error.

andsel:
I try to explain the behaviour, probably you could come out with more meaningful description.

`size_file` rules when we have to split the file, to create various segments. This plugin accumulate data in a locally temporary file, and then when it reach the `file_size` it split, creating a segment and then it's able to push that segment ion the remote S3.
Could be that for same tag (it's not clear to me what tag means, I imagine some common prefix) you have more split parts to upload.

`time_file` also rules when to split, but this time it consider the creation time. If the file was created 60 minutes, for example, split it.

The mixed behaviour time_and_size means:  once one of the two rules match, then split it

docs/index.asciidoc
andsel: Nice, sound really better then before, just one note
Err: reaise
Corridge raise

karenzone:
Thanks for the great explanation!  I think you uncovered the piece I was missing. 

Are  **these** statements correct (not to go in docs "as is", but so I understand):

If you set `rotation_strategy` to `size`, it uses the value in `size_file`. 
If you set `rotation_strategy` to `time`, it uses the value in `time_file`.
If you set `rotation_strategy` to `size_and_time`, it uses the values from `size_file` and `time_file`, and splits the file when either one matches.

andsel:
They are correct, exactly what the code do

docs/index.asciidoc
karenzone:
```suggestion
 If <<plugins-{type}s-{plugin}-rotation_strategy,`rotation_strategy`>> is set to `time` or `size_and_time`, then `time_file` cannot be set to 0. 
 Otherwise, the plugin raises a configuration error.
```

karenzone: Since you like the wording, I put it in as a suggestion with a link back to `rotation_strategy`.  I removed the "Some settings are not compatible" part after you helped me get a better understanding.  

Please take a look at Line 376 in the original file and Line 377 in your changes.  It looks like is still in the original file, but also repeated in your changes.

andsel: Thanks a lot @karenzone for the good reword, I integrated into the PR

PR #224, opened 2020-05-22
Bump to v.4.3.2 to pick up doc changes

karenzone:
Bump version to pick up doc changes from #223


karenzone:
Note to self @kaarenzone: Publishing pending clean CI

Issue #145, opened 2017-05-17
Temporary directory should be locked to one s3 output.


marcinc:
This issue is related to https://github.com/logstash-plugins/logstash-output-s3/issues/143 and discussion in https://github.com/logstash-plugins/logstash-output-s3/pull/144.

Multiple s3 outputs with `restore` enabled and default `temporary_directory` can lead to a race condition in crash recovery scenario whereby one s3 output picks up, uploads and removes outstanding temporary files from disk and remaining s3 outputs then try to do the same ending up not finding expected files on disk.

Proposed solution: Lock each s3 output to unique temporary directory.

rifelpet:
An additional side effect of sharing temp directories with multiple s3 output plugin instances is that if the instances are configured to log to different buckets, whichever instance has `restore` enabled will upload all the lingering files to its own bucket rather than the bucket associated with the plugin instance that created the file. This means log files can be uploaded to the wrong bucket.

Our fix is to just explicitly set the `temporary_directory` to something unique per instance and enable `restore` on all plugins, contrary to what the [comments suggest](https://github.com/logstash-plugins/logstash-output-s3/blob/9f5e2f2941b1731044a66d845bd672479c4236e2/lib/logstash/outputs/s3.rb#L122).

Issue #139, opened 2017-04-25
Better documentation for restore functionality


afn:
The `restore` option is poorly documented; it's hard to understand what this means:

```
  ## IMPORTANT: if you use multiple instance of s3, you should specify on one of them the "restore=> true" and on the others "restore => false".                                   
  ## This is hack for not destroy the new files after restoring the initial files.
  ## If you do not specify "restore => true" when logstash crashes or is restarted, the files are not sent into the bucket,
  ## for example if you have single Instance.
```

I'd love to help rewrite it but I'm not entirely clear on what it's *supposed* to say. :)

What's the "hack"? What happens if you set `restore => true` on multiple `s3` directives? Is this true even if they have different temporary directories?

Why can't we just default this option to true?

PR #236, opened 2021-06-17
Docs: more documentation on restore + temp dir

kares:
this is mostly documentation trying to clarify `restore => true` with mutliple outputs and their `temporary_directory`.

few minor refactorings as I was following the code base.

for context what **wasn't clear** (due code comments) is that we do support recovery with multiple outputs, if set right.

---

existing issues: https://github.com/logstash-plugins/logstash-output-s3/issues/145 https://github.com/logstash-plugins/logstash-output-s3/issues/139

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -395,7 +395,7 @@ Set the directory where logstash will store the tmp files before sending it to S
 default to the current OS temporary directory in linux `/tmp/logstash`.
 
 WARNING: Using multiple S3 outputs with `restore => true` requires unique directories
-per output as all of the directory's contents gets processed and deleted upon recovery.
+per output as all of the directory's contents gets processed and deleted upon recovery, and may cause data loss
 For example, an output using `/tmp/s3` and a second configured with `/tmp/s3/sub` would
 cause issues (having temporary directories `/tmp/s3/sub1` and `/tmp/s3/sub2` is fine).
 
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -395,9 +395,9 @@ Set the directory where logstash will store the tmp files before sending it to S
 default to the current OS temporary directory in linux `/tmp/logstash`.
 
 WARNING: Using multiple S3 outputs with `restore => true` requires unique directories
-per output as all of the directory's contents gets processed and deleted upon recovery, and may cause data loss
+per output. All of the directory's contents are processed and deleted upon recovery, and shared or nested directories can cause data loss.
 For example, an output using `/tmp/s3` and a second configured with `/tmp/s3/sub` would
-cause issues (having temporary directories `/tmp/s3/sub1` and `/tmp/s3/sub2` is fine).
+cause issues. Having temporary directories `/tmp/s3/sub1` and `/tmp/s3/sub2` is fine.
 
 [id="plugins-{type}s-{plugin}-time_file"]
 ===== `time_file` 
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -32,6 +32,7 @@ You can specify where to save them using the `temporary_directory` option.
 
 IMPORTANT: For configurations containing multiple s3 outputs with the `restore`
 option enabled, each output should define its own `temporary_directory`.
+Shared or nested directories can cause data loss upon recovery.
 
 ===== Requirements
 
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -255,9 +255,9 @@ The AWS Region
 Used to enable recovery after crash/abnormal termination.
 Temporary log files will be recovered and uploaded.
 
-NOTE: If you're using multiple S3 outputs than make sure to always set
-<<plugins-{type}s-{plugin}-temporary_directory,`temporary_directory`>> to a
-unique directory, otherwise the recovery mechanism won't work correctly.
+NOTE: If you're using multiple S3 outputs, always set
+<<plugins-{type}s-{plugin}-temporary_directory>> to a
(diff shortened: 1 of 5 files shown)


lib/logstash/outputs/s3.rb
```diff
@@ -147,6 +148,9 @@ class LogStash::Outputs::S3 < LogStash::Outputs::Base
 
   # Set the directory where logstash will store the tmp files before sending it to S3
   # default to the current OS temporary directory in linux /tmp/logstash
+  #
+  # NOTE: the reason we do not have a unique (isolated) temporary directory as a default, to support multiple plugin instances,
+  # is that we would have to rely on something static that does not change between restarts (e.g. a user set id => ...).
   config :temporary_directory, :validate => :string, :default => File.join(Dir.tmpdir, "logstash")
```
kares, line 154: was thinking of ways to warn users if there are multiple outputs running against the same dir (with `restore => true`).

the only thing I could think of would be to keep a `/tmp/logstash/my-tmp-dir/.logstash-output-s3` file around (deleted upon plugin stop) and than checking if there's another such file on start or in the parent dirs (e.g. `/tmp/logstash/.logstash-output-s3`) if so warning that restore won't work as expected ... but it feels a bit too much, thoughts?

robbavey: It's a tricky one because it's racy - we have no guarantee of the order of the plugin output creation, so the plugin attempting to write to the child folder could be created first - ie `/tmp/logstash/sub` can be created before `/tmp/logstash/`, so this test wouldn't flag a warning in this case.

Let's create a separate issue for this problem, and keep this one for the doc changes.

docs/index.asciidoc
```diff
@@ -30,8 +30,8 @@ Other S3 compatible storage solutions are not supported.
 S3 outputs create temporary files into the OS' temporary directory. 
 You can specify where to save them using the `temporary_directory` option.
 
-IMPORTANT: For configurations containing multiple s3 outputs with the restore
-option enabled, each output should define its own 'temporary_directory'.
+IMPORTANT: For configurations containing multiple s3 outputs with the `restore`
+option enabled, each output should define its own `temporary_directory`.
```
robbavey, line 34: Given the dangers, we should probably stress here that temporary directories must not be sub-folders of other outputs temporary directories otherwise data loss could occur

docs/index.asciidoc
```diff
@@ -388,7 +392,12 @@ Defaults to STANDARD.
   * Default value is `"/tmp/logstash"`
 
 Set the directory where logstash will store the tmp files before sending it to S3
-default to the current OS temporary directory in linux /tmp/logstash
+default to the current OS temporary directory in linux `/tmp/logstash`.
+
+WARNING: Using multiple S3 outputs with `restore => true` requires unique directories
+per output as all of the directory's contents gets processed and deleted upon recovery.
```
robbavey, line 398:
```suggestion
per output as all of the directory's contents gets processed and deleted upon recovery, and may cause data loss
```

robbavey (commented):
A few comments to emphasize the severity of the symptoms when mis-configuring temporary directories

robbavey (approved):
I'm good with the content, @karenzone would you mind giving it a final review?

docs/index.asciidoc
```diff
@@ -388,7 +392,12 @@ Defaults to STANDARD.
   * Default value is `"/tmp/logstash"`
 
 Set the directory where logstash will store the tmp files before sending it to S3
-default to the current OS temporary directory in linux /tmp/logstash
+default to the current OS temporary directory in linux `/tmp/logstash`.
+
+WARNING: Using multiple S3 outputs with `restore => true` requires unique directories
+per output as all of the directory's contents gets processed and deleted upon recovery, and may cause data loss
+For example, an output using `/tmp/s3` and a second configured with `/tmp/s3/sub` would
+cause issues (having temporary directories `/tmp/s3/sub1` and `/tmp/s3/sub2` is fine).
```
karenzone, line 400:
```suggestion
default to the current OS temporary directory in linux `/tmp/logstash`.

WARNING: Using multiple S3 outputs with `restore => true` requires unique directories
per output. All of the directory's contents are processed and deleted upon recovery, and shared or nested directories can cause data loss.
For example, an output using `/tmp/s3` and a second configured with `/tmp/s3/sub` would
cause issues. Having temporary directories `/tmp/s3/sub1` and `/tmp/s3/sub2` is fine.
```

karenzone: Minor rewording suggestion for your consideration.

docs/index.asciidoc
```diff
@@ -255,6 +255,10 @@ The AWS Region
 Used to enable recovery after crash/abnormal termination.
 Temporary log files will be recovered and uploaded.
 
+NOTE: If you're using multiple S3 outputs than make sure to always set
+<<plugins-{type}s-{plugin}-temporary_directory,`temporary_directory`>> to a
+unique directory, otherwise the recovery mechanism won't work correctly.
```
karenzone, line 260:
```suggestion
NOTE: If you're using multiple S3 outputs, always set
<<plugins-{type}s-{plugin}-temporary_directory>> to a
unique directory. Otherwise the recovery mechanism won't work correctly.
```

karenzone: Tip: You don't have to add link override text if you want the link to use the heading. That is, `<<plugins-{type}s-{plugin}-temporary_directory>> links to `temporary_directory` so no override is needed.

karenzone (commented):
Minor suggestions for your consideration.  Otherwise, build cleanly and LGTM

PR #248, opened 2022-06-16
Refactor: avoid usage of CHM (JRuby 9.3.4 work-around)

kares:
using `.forEach` will trigger a bug with JRuby 9.3.4 (on second execution of the call-site)

since we also hit some other issues using ConcurrentHashMap from ruby I went ahead and refactored to `Concurrent::Map` which is pretty much the same thing.


andsel (approved):
LGTM

PR #249, opened 2022-06-21
Initial setup of Java-Ruby mixed environment, Gzip recover Java class…

mashhurs:
## Issue description
When using `GZIP` encoding option with output to AWS S3 plugin, there are cases where Logstash may be crashed. When Logstash crashed GZIP stream is left opened and no tail in the file exist. Logstash uploads corrupted file to S3 at restart but customers who download S3 file and use, they figured out the file is corrupted.
This PR aims to recover the corrupted file at restart time and upload healthy GZIP file to S3.

_FYI: Requested the case where/when/how Logstash crashed, I will investigate once I get response_

## Acceptance Criteria
Logstash should always upload healthy GZIP/plain text files to AWS S3.

## Solution explanation
Look at the note section for details: option **4 is recommended and implemented.**

### Unit testing
- `GzipUtilTest` unit test class covers to recover, compress and decompress success and failure cases accordingly.

### E2E testing
- After pulling this change and setup local dev with Logstash-code, setup your AWS S3 bucket with file uploadable policy.
- Use the S3 output setting, eg.:
```
input {
    stdin { }
}
output {
    s3 {
        region => "region"
        bucket => "logstash-mashhur-test"
        codec => "json_lines"
        canned_acl => "private"
        prefix => "test-%{+YYYY.MM.dd}"
        additional_settings => {
            "force_path_style" => true
        }
        upload_queue_size => 10
        upload_workers_count => 2
        size_file => 1024
        rotation_strategy => "size"
        encoding => gzip
        temporary_directory => "/path/to/local/temp/s3"
        validate_credentials_on_root_bucket => false
    }

    stdout {}
}
```
- Run Logstash and type on the console, kill Logstash with `kill -9 PID` where PID can be fetched with `ps -la | grep logstash`
- Re-run Logstash and see the S3 bucket if it uploaded, you may seen on console that Logstash warns recovered file.
```
[2022-06-22T15:38:43,062][WARN ][org.logstash.plugins.outputs.s3.GzipUtil][main] Corrupted file recovered, path:/path/of/file-recovered.txt
```
- Download S3 object with `aws s3 cp s3://logstash-mashhur-test/path-to-file.txt.gz /local/path/to/download/`
- Unzip file with `gunzip downloaded-file.txt.gz` and open the file.

## Additional Notes
Discussed solutions:

1. [Gzip recovery] Pull out the GZIP header and decompress healthy blocks till we find a tail in Plain Ruby.
   - Ruby GZIP interfaces (`GzipReader`, `GzipWriter`) do not allow to access GZIP blocks/headers byte-by-byte. They always validate the file on any read actions. - NOT FEASIBLE.

2. Logstash uses plain text and when uploading to S3 it zips the file. 
   - For this case, the size policy will not work properly. Saying that, when users set `output -> s3 -> size_file` users' intention is ZIP file size not ongoing plain text. Since Logstash looses accuracy, this option would be a low priority.

3. External tool
  - Existing tools such as `gz-recover` will work as expected and this is an OSS. However, for the long term (eg. if security vulnerabilities found) this option would cost high since we may need to contribute or because of poor maintenance. - low priority.

4. [Gzip recovery - RECOMMENDED, IMPLEMENTED] Pull out the GZIP header and decompress healthy blocks till we find a tail in Java.
  - Java has quite good input stream interfaces to access each file bytes. Initially, read the file and pulled out the header but `GZIPInputStream` interface does same whenever stream starts. That means we don't necessarily look up for header or tail operations. Instead, capturing `Unexpected end of ZLIB input stream` exception satisfies our condition.
  - However, this solution requires the plugin environment to be a dynamic: Ruby-Java mixed. We have plugins (`input-beats`, `input-http`) already took initiative to move to such environment. So, this change also includes Java-Ruby mixin env setups.


lib/logstash/outputs/s3/temporary_file.rb
```diff
@@ -61,10 +69,23 @@ def empty?
         def self.create_from_existing_file(file_path, temporary_folder)
           key_parts = Pathname.new(file_path).relative_path_from(temporary_folder).to_s.split(::File::SEPARATOR)
 
+          # recover gzip file and compress back before uploading to S3
+          if file_path.end_with?(GZIP_EXTENSION)
```
mashhurs, line 73: `@encoding` option might change when Logstash restarts, file extension check seems accurate.

logstash-output-s3.gemspec
```diff
@@ -1,13 +1,13 @@
 Gem::Specification.new do |s|
   s.name            = 'logstash-output-s3'
-  s.version         = '4.3.6'
+  s.version         = '4.3.7'
```
mashhurs, line 3: @robbavey do you think we should change major version? IMO we are fixing the corrupted upload issue so changed minor version. But this change also includes Java-Ruby mixed environment support. Confusing since setup changes does not give any benefits. Need your feedback.

logstash-output-s3.gemspec
```diff
@@ -1,13 +1,13 @@
...
   s.licenses        = ['Apache-2.0']
   s.summary         = "Sends Logstash events to the Amazon Simple Storage Service"
   s.description     = "This gem is a Logstash plugin required to be installed on top of the Logstash core pipeline using $LS_HOME/bin/logstash-plugin install gemname. This gem is not a stand-alone program"
   s.authors         = ["Elastic"]
   s.email           = 'info@elastic.co'
   s.homepage        = "http://www.elastic.co/guide/en/logstash/current/index.html"
-  s.require_paths = ["lib"]
+  s.require_paths = ["lib", "vendor/jar-dependencies"]
```
mashhurs, line 10: @robbavey thanks so much for letting me know this!

build.gradle
```diff
@@ -0,0 +1,83 @@
...
+    mavenCentral()
+}
+
+dependencies {
+    testImplementation 'junit:junit:4.12'
+    testImplementation 'org.apache.logging.log4j:log4j-core:2.17.0'
+    implementation "com.fasterxml.jackson.core:jackson-core:${jacksonVersion}"
+    implementation "com.fasterxml.jackson.core:jackson-annotations:${jacksonVersion}"
+    implementation "com.fasterxml.jackson.core:jackson-databind:${jacksonDatabindVersion}"
+    implementation "com.fasterxml.jackson.module:jackson-module-afterburner:${jacksonVersion}"
```
kares, line 29: :scissors: do not think these jackson dependencies are needed

lib/logstash/outputs/s3.rb
```diff
@@ -398,12 +398,14 @@ def restore_from_crash
 
     temp_folder_path = Pathname.new(@temporary_directory)
     Dir.glob(::File.join(@temporary_directory, "**/*"))
-      .select { |file| ::File.file?(file) }
-      .each do |file|
-      temp_file = TemporaryFile.create_from_existing_file(file, temp_folder_path)
+      .select { |file_path| ::File.file?(file_path) }
```
robbavey, line 401: We will need to figure out a way between the recovery code and this glob to ignore "recovery in progress" - currently, in the event of a failure/crash during "recovery in progress", this `glob` definition will pick up the original file to be recovered, plus any files in mid-recovery - ie `ls.s3<>part0.txt.gz`, `ls.s3<>.part0-recovered.txt.gz`, and `ls.s3<>.part0-recovered.txt.gz`. This can lead to a few different outcomes - duplicate data being uploaded to S3, potentially multiple `-recovered.txt` and `-recovered.txt.gz` extensions, and, in the event one of these files is empty, `clean_temporary_file(temp_file)` will be called, which deletes the entire sub directory that all of these files are stored in.

src/main/java/org/logstash/plugins/outputs/s3/GzipUtil.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Decompresses and recovers corrupted GZIP file into target file.
+     * @param sourcePath a path where corrupted GZIP is located
+     * @param targetPath a path where target can be placed
+     * Note: raises an IO exception other than GZIP dead blocks.
+     */
+    public static void recover(String sourcePath, String targetPath) {
+        try {
+            GzipUtil.decompress(Paths.get(sourcePath), Paths.get(targetPath));
```
robbavey, line 48: If the recovery returns `0` bytes, then we should probably skip the rest of the recovery process, to avoid uploading an empty zip file

src/main/java/org/logstash/plugins/outputs/s3/GzipUtil.java
```diff
@@ -0,0 +1,96 @@
...
+                logger.warn("Corrupted file recovered, path:" + targetPath);
+            }
+        }
+    }
+
+    /**
+     * Decompresses a given compressed source file into target file.
+     * @param sourcePath a path where compressed source is located
+     * @param targetPath a path where target can be placed
+     * Note: raises an IO exception if any break happens.
```
robbavey, line 63: Same as previous - javadoc states "raise", but will log instead

src/main/java/org/logstash/plugins/outputs/s3/GzipUtil.java
```diff
@@ -0,0 +1,96 @@
...
+     */
+    public static void decompress(String sourcePath, String targetPath) {
+        try {
+            GzipUtil.decompress(Paths.get(sourcePath), Paths.get(targetPath));
+        } catch (IOException exception) {
+            logger.error("Error occurred while decompressing the file, error={}", exception.getMessage());
```
robbavey, line 69: Same comment as per `recover`

robbavey (changes requested):
First pass through

src/main/java/org/logstash/plugins/outputs/s3/GzipUtil.java
```diff
@@ -0,0 +1,96 @@
...
+        } catch (IOException exception) {
+            logger.error("Error occurred while compressing the file, error={}", exception.getMessage());
+        }
+    }
+
+    /**
+     * Decompresses and recovers corrupted GZIP file into target file.
+     * @param sourcePath a path where corrupted GZIP is located
+     * @param targetPath a path where target can be placed
+     * Note: raises an IO exception other than GZIP dead blocks.
```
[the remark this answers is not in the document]
mashhurs: Fixed.

src/main/java/org/logstash/plugins/outputs/s3/GzipUtil.java
```diff
@@ -0,0 +1,96 @@
...
+     */
+    public static void recover(String sourcePath, String targetPath) {
+        try {
+            GzipUtil.decompress(Paths.get(sourcePath), Paths.get(targetPath));
+        } catch (IOException exception) {
+            // raise an exception if expected exception is not end of ZLIB related.
+            if (CORRUPTED_FILE_ERROR.equals(exception.getMessage()) == false) {
+                logger.error("Error occurred while compressing the file, error={}", exception.getMessage());
```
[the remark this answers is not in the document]
mashhurs: Well, so far the only case when GZIP might be corrupted is `no tail` case where stream is left opened when LS is crashed. In case if other cases happen, currently, LS will leave file as it is, users will be able to find it under temporary path. No any delete action will be proceeded. And also, most importantly, we don't damage/crash LS and let it continue.
Let me know your thought if we can improve the situation.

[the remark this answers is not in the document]
robbavey: Unless, I'm missing something, I don't think LS will leave the file as it is. With this change, the error will be logged, but processing will continue as normal - uploading the corrupt file to S3, and deleting the file from local storage.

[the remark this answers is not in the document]
mashhurs: File path changes when corrupted file is recovered (`file_path = self.recover(file_path)`) and new temp file will be returned (with `create_from_existing_file`). If LS does not recover the file its size is still zero (in temporary file) and it does not remote the file ([source](https://github.com/mashhurs/logstash-output-s3/blob/60ba59e392f2e515c2d2ad2fd6d16dded1636569/lib/logstash/outputs/s3.rb#L404))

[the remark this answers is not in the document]
robbavey: If the file size is 0, the temporary file _and_ the original "corrupt" file will be deleted - the process of deleting a "temporary file" also [deletes the folder](https://github.com/logstash-plugins/logstash-output-s3/blob/main/lib/logstash/outputs/s3/temporary_file.rb#L52) it is a part of.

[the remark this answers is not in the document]
mashhurs: Oh it is because of temp folder... yeah agree that temp folder will be removed. Thanks for letting know, reproducing such case is not easy. How do you think if we replace following logic 

```
// s3.rb:404L
        if temp_file.size > 0
        @logger.debug? && @logger.debug("Recovering from crash and uploading", :path => temp_file.path)
        @crash_uploader.upload_async(temp_file,
                                     :on_complete => method(:clean_temporary_file),
                                     :upload_options => upload_options)
      end
      if @encoding != "gzip" && temp_file.size != 0
        clean_temporary_file(temp_file)
      end
```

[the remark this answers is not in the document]
mashhurs: But I don't see the logic where deleting the folder... Can you plz let me know?
For any unsuccessful conditions, In `clean_temporary_file(temp_file)` new file for recovery (not original corrupted ZIP) will be passed and removed which is okay.

src/main/java/org/logstash/plugins/outputs/s3/GzipUtil.java
mashhurs: This will covered by [`s.rb:L405~L413`](https://github.com/logstash-plugins/logstash-output-s3/pull/249/files#diff-bdae3d005e86be10100b9ba58d128b55d13c8b8bb4e4ffd43f2c0956c22666e2R405-R412)

lib/logstash/outputs/s3.rb
mashhurs: Is `Dir.glob().select{}` async for this case?
The possible cases would be 
- multiple LS instances are accessing to same S3 temp path but as we discussed offline the recommendation would be providing a separate path for each instance.
- unrecoverable zip files which always stay

So, I have added a logic where figures out the under recovery process files (`under_recovery_files`) and skips the file if it is in this list.

src/main/java/org/logstash/plugins/outputs/s3/GzipUtil.java
robbavey: It is possible for an "empty" gzip file to be created which includes the header but no content, which I don't think will be covered by the above case, and will create an empty recovery file and upload a `gzip` version.

(I could make this happen with `input { generator { count => 250 }}`)

lib/logstash/outputs/s3.rb
```diff
@@ -397,16 +398,43 @@ def restore_from_crash
...
+      if file_path.include?(TemporaryFile.recovery_file_name_tag)
+        skip_files << file_path
+        if file_path.include?(TemporaryFile.gzip_extension)
+          # also include the original corrupted gzip file
+          skip_files << file_path.gsub(TemporaryFile.recovery_file_name_tag, "")
+        end
       end
     end
+    skip_files
   end
```
robbavey, line 439: Somewhat of an edge case, but if we skip the original zip file, if a failure/crash occurs during a valid recovery, the directory will be in a state of 'limbo' where the contents of the folder will be neither deleted nor uploaded.

Not including the original zip file in this list would help recovery in this case, but will likely require some additional logic to handle unrecoverable errors (maybe an indication that the file is "damaged"), and remove recovery that is half finished?

```suggestion
  def get_under_recovery_files(files)
    files.select { |file_path| file_path.include?(TemporaryFile.recovery_file_name_tag)}
 ```

File: src/main/java/org/logstash/plugins/outputs/s3/GzipUtil.java (hunk 6 above)
Comment (mashhurs) [reply src/main/java/org/logstash/plugins/outputs/s3/GzipUtil.java:48]: Thanks, confirmed/tested possible empty zip cases and updated the PR.

File: lib/logstash/outputs/s3.rb (hunk 11 above)
Comment (mashhurs) [reply lib/logstash/outputs/s3.rb:439]: The only possible case I can see is when S3 is unavailable (or LS is having an issue with S3). We are not removing the temporary folder unless we achieve the successful completion of recovery. And I think it is better to get know the cases better first.

Also, suggested logic will not work because as you can see L431 & L434 lines are adding two entries where suggestion will add only one.

File: lib/logstash/outputs/s3/temporary_file.rb [hunk 12]
```diff
@@ -61,10 +69,38 @@ def empty?
         def self.create_from_existing_file(file_path, temporary_folder)
           key_parts = Pathname.new(file_path).relative_path_from(temporary_folder).to_s.split(::File::SEPARATOR)
 
+          # recover gzip file and compress back before uploading to S3
+          if file_path.end_with?("." + GZIP_EXTENSION)
+            file_path = self.recover(file_path)
+          end
           TemporaryFile.new(key_parts.slice(1, key_parts.size).join("/"),
-                         ::File.open(file_path, "r"),
+                         ::File.exist?(file_path) ? ::File.open(file_path, "r") : nil, # for the nil case, file size will be 0 and upload will be ignored.
```
robbavey, line 77: Supplying a `nil` file descriptor here will cause the `size` method to raise rather than return `0`

lib/logstash/outputs/s3.rb
robbavey: Yes, that was the scenario that I was testing out. My main concern with the current approach is that we are skipping over files in the temporary directory without informing the user - I think we should either log that we are skipping over those files, or attempt to re-upload the zip file (logging that too)

(my suggested code snippet was to include the original zip file to attempt to re-upload it, rather than a like for like swap of your original code)

robbavey: I'm ok with logging - I'd much rather that than silently skipping over. We can iterate over this in a future PR

lib/logstash/outputs/s3/temporary_file.rb
mashhurs: Oh sorry, missed to update the `TemporaryFile` class, now included.

lib/logstash/outputs/s3.rb
```diff
@@ -397,16 +398,48 @@ def restore_from_crash
...
-        clean_temporary_file(temp_file)
+        temp_file = TemporaryFile.create_from_existing_file(file_path, temp_folder_path)
+        if temp_file.size > 0
+          @logger.debug? && @logger.debug("Recovering from crash and uploading", :path => temp_file.path)
+          @crash_uploader.upload_async(temp_file,
+                                       :on_complete => method(:clean_temporary_file),
+                                       :upload_options => upload_options)
+        end
+        # do not remove if Logstash tries to recover but fails
+        if @encoding != GZIP_ENCODING && temp_file.size != 0
```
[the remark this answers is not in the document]
robbavey: I think it might be worthwhile to see some tests here - what happens with:

- standard plain text recovery
- standard gzip recovery
- empty plain text recovery
- empty gzip recovery

robbavey (changes requested):

lib/logstash/outputs/s3.rb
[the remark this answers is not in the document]
mashhurs: It seems default (`none` or `plain txt`) encoding cases are covered in the integration test, [`restore_from_crash_spec.rb`](https://github.com/logstash-plugins/logstash-output-s3/blob/d96531b573a8ff4e2cc216ad1c8cb83027000499/spec/integration/restore_from_crash_spec.rb) and I have added the cases where Logstash restores the file, uploads and removes locally.

- Run integration test, as we synced offline, make sure to set the AWS ENV params including `AWS_SESSION_TOKEN`:
`bundle exec rspec logstash-plugins/logstash-output-s3/spec/integration/restore_from_crash_spec.rb:69 -t integration:true`
- Result logs:
```
Finished in 1 minute 22.39 seconds (files took 9.97 seconds to load)
3 examples, 0 failures
```

lib/logstash/outputs/s3/temporary_file.rb
mashhurs: yeah, I double checked that forward-able methods will not be called when restoration fails. In the future, it would be nice if I can find a strategy to figure out recovery status rather than checking file existence.

robbavey: Is there a benefit to returning a `TemporaryFile` with a `nil` `fd` over returning `nil` here?

mashhurs: Recovered file will not exist if recovery fails or restored txt file is empty. That leads to bring a failure (no file exist exception as I remember) when checking a size.

xiaomaisuii:
Hello, I saw that the fix code #249 has been merged into the logstash-plugins/logstash-integration-aws repository. Could you let me know when this fix will be merged into the logstash-plugins/logstash-output-s3 repository, or if it has already been fixed in a particular version of logstash?

PR #250, opened 2022-07-19
Bump plugin gem version.

mashhurs:
Bumping version which had to be included in #249


robbavey:
Can we make this a `minor` update? That would also require changes to the [version file](https://github.com/logstash-plugins/logstash-output-s3/blob/main/VERSION) and the [changelog](https://github.com/logstash-plugins/logstash-output-s3/blob/main/CHANGELOG.md)

robbavey (approved):
LGTM

PR #251, opened 2022-12-21
Fixes the no such file or directory issue.

mashhurs:
## Issue description
When temporary file is staled and Logstash is closing the s3-output plugin, the plugin is accessing the stolen file and producing an error:

```
[2022-12-21T17:05:58,371][ERROR][logstash.javapipeline    ][main] Pipeline error {:pipeline_id=>"main", 
:exception=>#<Errno::ENOENT: No such file or directory - No such file or directory -
 /temp/s3-temp/4e46daa5-07af-4085-a819-bac9ced3d25a/test-2022.12.21/ls.s3.9b6828b5-cff3-4b97-b1e1-ff83af405817.2022-12-21T17.05.part0.txt>, :backtrace=>["org/jruby/RubyFileTest.java:248:in `size'", 
"/temp/logstash-output-s3/lib/logstash/outputs/s3/temporary_file.rb:50:in `size'",
 "/temp/logstash-output-s3/lib/logstash/outputs/s3.rb:373:in `upload_file'", "/temp/logstash-output-s3/lib/logstash/outputs/s3.rb:276:in `block in close'", "/temp/logstash-output-s3/lib/logstash/outputs/s3/file_repository.rb:79:in `block in each_files'", 
[...]

PR #252, opened 2022-12-21
Stale cleanup race condition

yaauie:
Alternative to #251 that catches an additional race-condition

 - Fixes several closely-related race conditions that could cause plugin crashes or data-loss
    - race condition in initializing a prefix could cause one or more local temp files to be abandoned and only recovered after next pipeline start (replace `Concurrent::Map#fetch_or_store` with _atomic_ `Concurrent::Map#computeIfAbsent`)
    - race condition in stale watcher could cause the plugin to crash when working with a stale (empty) file that had been deleted (mark `FileRepository::PrefixedValue` as deleted while we have exclusive access to it, and avoid using a marked-deleted `FileRepository::PrefixedValue` throughout)
    - race condition in stale watcher could cause a non-empty file to be deleted if bytes were written to it after it was detected as stale (check-and-delete within a single exclusive access, which requires that `FileRepository::PrefixedValue#with_lock` use reentrant `Monitor`)

***

This will need to be forward-ported to the integrated version of this plugin -> `logstash-integration-aws`

### DIFF
--- a/lib/logstash/outputs/s3/file_repository.rb
+++ b/lib/logstash/outputs/s3/file_repository.rb
@@ -15,8 +15,9 @@ module LogStash
         class PrefixedValue
           def initialize(file_factory, stale_time)
             @file_factory = file_factory
-            @lock = Mutex.new
+            @lock = Monitor.new
             @stale_time = stale_time
+            @is_deleted = false
           end
 
           def with_lock
@@ -34,7 +35,14 @@ module LogStash
           end
 
           def delete!
-            with_lock{ |factory| factory.current.delete! }
+            with_lock do |factory|
+              factory.current.delete!
+              @is_deleted = true
+            end
+          end
+
+          def deleted?
+            with_lock { |_| @is_deleted }
           end
         end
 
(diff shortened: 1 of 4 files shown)


lib/logstash/outputs/s3/file_repository.rb
```diff
@@ -95,18 +138,31 @@ def size
...
-            v.delete!
+        def remove_if_stale(prefix_key)
+          # we use the ATOMIC `Concurrent::Map#compute_if_present` to atomically
+          # detect the staleness, mark a stale prefixed factory as deleted, and delete from the map.
+          @prefixed_factories.compute_if_present(prefix_key) do |prefixed_factory|
+            # once we have retrieved an instance, we acquire exclusive access to it
+            # for stale detection, marking it as deleted before releasing the lock
+            # and causing it to become deleted from the map.
+            prefixed_factory.with_lock do |_|
+              if prefixed_factory.stale?
```
mashhurs, line 149: we are trying to lock the prefixed_factory in `stale?` as well, isn't deadlock?

lib/logstash/outputs/s3/file_repository.rb
```diff
@@ -95,18 +138,31 @@ def size
...
+                prefixed_factory.delete! # mark deleted to prevent reuse
+                nil # cause deletion
```
mashhurs, line 151: Can I know what is the purpose of returning value?

lib/logstash/outputs/s3/file_repository.rb
```diff
@@ -72,17 +80,52 @@ def keys
...
+        # @return [Object]: the value returned by the provided block
         def get_factory(prefix_key)
-          prefix_val = @prefixed_factories.fetch_or_store(prefix_key) { @factory_initializer.create_value(prefix_key) }
+
+          # fast-path: if factory exists and is not deleted, yield it with exclusive access and return
+          prefix_val = @prefixed_factories.get(prefix_key)
+          prefix_val&.with_lock do |factory|
+            # intentional local-jump to ensure deletion detection
+            # is done inside the exclusive access.
+            return yield(factory) unless prefix_val.deleted?
```
mashhurs, line 102: I was thinking to use global lock for the works case but I think this is better approach and eliminates [the scenario you mentioned](https://github.com/logstash-plugins/logstash-output-s3/pull/251#pullrequestreview-1226465674).  Thanks!

lib/logstash/outputs/s3/file_repository.rb
yaauie: We already have the (reentrant) lock, so incrementing our holds on it briefly won't ever have lock contention.

lib/logstash/outputs/s3/file_repository.rb
yaauie: It is a part of the `Concurrent::Map#compute_if_present` contract; the result of the block will be stored, or if `nil` will cause the value to be deleted.

mashhurs (approved):
LGTM!

mashhurs (approved):
LGTM!

